PR #1, opened 2023-07-30
init project

hijkzzz:


openllmai (approved):
add deepspeedchat for ref and init openllama2

PR #64, opened 2023-08-18
updata readme

pikaqqqqqq:
add installation for deepspeed


hijkzzz:
Hi, we just need `python3 -m pip install .`
then deepspeed will be installed~

hijkzzz:
maybe you can provide some new install instrucions for non-nvidia-docker users.
thx

PR #76, opened 2023-08-21
support dataset with subfold

wwxFromTju:
Support for specifying the subfold of the dataset via @, e.g. 'Dahoas/full-hh-rlhf,Anthropic/hh-rlhf,Anthropic/hh-rlhf@harmless-base'


wwxFromTju:
@hijkzzz

hijkzzz (approved):
a good idea

PR #88, opened 2023-08-26
feat: add wandb logger in ppo trainer

dabney777:


hijkzzz:
Hi, I have added this fucntion.

hijkzzz:
maybe you could help to add some text log features like, thx
image

PR #132, opened 2023-11-05
fix log/eval/save steps

catqaq:


hijkzzz:
There is no need to do anything special here. eval in the process of train will not cause any serious consequences

PR #135, opened 2023-11-13
fix: remove unnecessary allreduce which will cause timeout

wuxibin89:
This all_reduce will cause timeout on ray because logs_dict is asymmetric across all actors. When train on ray, some actors are responsible for triggering remote critic model training


hijkzzz:
That's an interesting question: 
This all reduce is necessary for global statistics,
But it also cause a timeout in the ray environment

hijkzzz:
I think we can ignore these two special variables ("critic_loss","values") in all reduce+ray

wuxibin89:
@hijkzzz status is all_reduce in training loop

PR #140, opened 2023-11-18
fix examples/scripts/train_sft_llama.sh

xffxff:
While executing the script `train_sft_llama.sh`, I encountered an error 
```
train_sft_llama.sh: line 29: warning: here-document at line 3 delimited by end-of-file (wanted `EOF')
+ read -r -d '' training_commands
```

After some investigation, I found that removing the `\` before 'EOF' solved the problem, it might be a typo


hijkzzz:
thanks for your contribution

PR #153, opened 2023-12-11
Repair parquet file reading issue in utils.py

tsaoyu:
The dataset complains `Unsupported file types: parquet` which is misleading.

Also https://github.com/OpenLLMAI/OpenRLHF/compare/main...tsaoyu:OpenRLHF:patch-1#diff-6e5df94c43d3630b7bab08546e9fb0161793683ec823ab627e3858783907e407L144 looks very suspicious to me.


catqaq (approved):
I will merge this change first, and then do unified processing for local file support.

tsaoyu:
sure

PR #162, opened 2023-12-16
Optimize reward score gather/scatter

li-plus:
Similar to #152, speed up reward score gathering/scattering by eliminating nested python loops.


catqaq:
thx!
We will review this change as quickly as possible. After the first official version, we plan to conduct a round of code review to enhance code quality and identify potential areas for performance improvement. However, due to limited manpower, this process won't be very fast.

PR #165, opened 2023-12-19
Optimize padding removal

li-plus:
Changes:
* eliminate the inner python loop
* don't offload replay buffer since it's small

Main branch `remove_padding_in_sequences` costs 169ms
image

With this PR it just costs 2ms on cpu. Corresponding cuda kernels are also launched and finished within this 2ms.
image


hijkzzz:
I agree that `eliminate the inner python loop` with `argmax`
But GPU memory is very sensitive to large model training, and we should save as much as possible.

li-plus:
It's a tradeoff between speed and memory. I suggest making it an option for offloading, maybe in a future PR. Anyway, I'll revert this change soon.

PR #166, opened 2023-12-20
Fix flash attention option

li-plus:
The main branch did not use flash attention in RLHF training at all, even if you specify `--flash_attn`.

`AutoConfig.from_pretrained` does not respect the `use_flash_attention_2` option. `AutoModelForCausalLM.from_config` or `AutoModelForCausalLM.from_pretrained` does.

### DIFF
--- a/openrlhf/models/actor.py
+++ b/openrlhf/models/actor.py
@@ -32,9 +32,12 @@ class Actor(nn.Module):
                     pretrain_or_model,
                     torch_dtype="auto",
                     trust_remote_code=True,
+                )
+                self.model = AutoModelForCausalLM.from_config(
+                    config,
+                    trust_remote_code=True,
                     use_flash_attention_2=use_flash_attention_2,
                 )
-                self.model = AutoModelForCausalLM.from_config(config, trust_remote_code=True)
             else:
                 self.model = AutoModelForCausalLM.from_pretrained(
                     pretrain_or_model,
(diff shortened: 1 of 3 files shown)


hijkzzz:
Hi, may I ask if you can point out the FlashAttention2 + .from_config description in the HF documentation? I looked for a while but couldn't find it, so I'm not sure what the right way to do this is.

wuxibin89:
`AutoConfig.from_pretrained` return architecture specific config, like LlamaConfig, which are derived from base class `PretrainedConfig` https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L49
There's no use_flash_attention_2 in this class, so I think we actually didn't enable flash attention correctly.

wuxibin89:
`AutoModelForCausalLM.from_config` has no keyword argument `use_flash_attention_2` either, maybe we should remove from_config completely.

li-plus:
> I just noticed that the option is configured in the config class: https://github.com/huggingface/transformers/blob/769a9542de4e8b23f0a551738e18760621f463e8/src/transformers/modeling_utils.py#L1299

Correct. I can confirm flash attn is enabled because flash kernels are launched in my trace. You may also double check.

hijkzzz:
Fixed in https://github.com/OpenLLMAI/OpenRLHF/commit/84917bafc1e02d5d1fc4de7eed982487a2f9ee9a
Refer to https://github.com/mosaicml/llm-foundry/blob/289536bd9403d8c188f8e968396475e92e2abdd2/llmfoundry/models/hf/hf_causal_lm.py#L105

PR #200, opened 2024-01-25
Enable overlap_comm for better performance

li-plus:
Overlap gradient allreduce and backward. Port my PR to ds-chat https://github.com/microsoft/DeepSpeedExamples/pull/846 here.

### DIFF
--- a/openrlhf/utils/deepspeed_utils.py
+++ b/openrlhf/utils/deepspeed_utils.py
@@ -23,6 +23,7 @@ def get_train_ds_config(
     device = "cpu" if offload else "none"
     zero_opt_dict = {
         "stage": stage,
+        "overlap_comm": True,
         "offload_param": {"device": device},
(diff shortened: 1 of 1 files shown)


wuxibin89 (approved):

hijkzzz:
This option should be turned on depending on the situation because overlap comm takes up a lot of extra GPU memory

wuxibin89:
> This option should be turned on depending on the situation because overlap comm takes up a lot of extra GPU memory

Well, may be we should expose some deepspeed configuration and let user passing them by command line arguments?

hijkzzz:
I have reverted this MR, since it causes the training unstable: https://wandb.ai/janhu9527/openrlhf_train_ppo?workspace=user-janhu9527

hijkzzz:
ee202110e00ce2391e1509c2cc7afe7

PR #213, opened 2024-02-18
fix: ignore non-persistent named buffer when save model

wuxibin89:


catqaq (commented):
Perhaps a list comprehension implementation would be cleaner?

Issue #217, opened 2024-02-21
`position_ids` related PPO bug


tianhao-nexusflow:
I'm working to reproduce results and align them with `trlx`. After careful code review, I've identified this issue:

Concern: [link](https://github.com/OpenLLMAI/OpenRLHF/blob/ecfb6e8704a4e7d0c82f1f1134a96aa3b2b62d2d/openrlhf/trainer/ppo_utils/experience_maker.py#L124), Precomputing log probabilities using only `token_ids` and `attention_mask` may not fully account for the model's use of positional embeddings. As for llama model, the same `token_ids` with or without left padding would have different result even it explicitly pass in the `attention_mask`. A common strategy to correct this is to explicitly pass in the `position_ids` as in [link](https://github.com/CarperAI/trlx/blob/3340c2f3a56d1d14fdd5f13ad575121fa26b6d92/trlx/trainer/accelerate_ppo_trainer.py#L180)

Potential Impact: The same padding related issue could also impact the generate quality, while I need more time to carefully look into the code. Will this also impact the critic and reward modules?

hijkzzz:
This issue is very interesting and I will confirm it as soon as possible. If so, the bug is caused by huggingface and we will fix it

hijkzzz:
fixed

PR #219, opened 2024-02-22
use_right_pad

hijkzzz:
#217 use right pad to reduce precision issues

### DIFF
--- a/openrlhf/trainer/ppo_utils/replay_buffer.py
+++ b/openrlhf/trainer/ppo_utils/replay_buffer.py
@@ -92,7 +92,7 @@ def make_experience_batch(items: List[BufferItem]) -> Experience:
     )
     for key in keys:
         vals = [getattr(item, key) for item in items]
-        batch_data = zero_pad_sequences(vals, "left")
+        batch_data = zero_pad_sequences(vals, "right")
         kwargs[key] = batch_data
 
     kwargs["info"] = {}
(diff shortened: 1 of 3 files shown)


hijkzzz:
for this MR, there are some bugs:

image

openrlhf/trainer/ppo_utils/replay_buffer.py
```diff
@@ -92,7 +92,7 @@ def make_experience_batch(items: List[BufferItem]) -> Experience:
     )
     for key in keys:
         vals = [getattr(item, key) for item in items]
-        batch_data = zero_pad_sequences(vals, "left")
+        batch_data = zero_pad_sequences(vals, "right")
```
wuxibin89, line 95: Should we make sequences right pad right after generation, since we have to calculate action_logits/ref_logits/values/rewards?
```bash
>>> attention_mask = torch.Tensor([[0,1,1,0], [1,1,1,1],[1,1,1,0]])
>>> attention_mask
tensor([[0., 1., 1., 0.],
        [1., 1., 1., 1.],
        [1., 1., 1., 0.]])
>>> attention_mask = torch.Tensor([[1,1,0,0], [1,1,1,1],[1,1,1,0]])
>>> attention_mask
tensor([[1., 1., 0., 0.],
        [1., 1., 1., 1.],
        [1., 1., 1., 0.]])
```

hijkzzz: It does use the right pad, but there are some unknown bugs

PR #231, opened 2024-03-04
Use existing wandb login if available.

mgerstgrasser:
Currently wandb logging is enabled by passing a wandb API key as a CLI argument. However, that makes it easy to accidentally commit an API key as part of a shell script. It is also not necessary, as wandb stores API keys systemwide after running `wandb login` on the command line. 

This PR makes a minimal change to only run `wandb.login()` in Python if the user is not already logged in. The effect is that you can just pass `--use_wandb True` (or any value) instead of `--use_wandb ....api_key...` if you're already logged in. However, if you are _not_ already logged in, this PR does not break things, and will still accept an API key through the CLI argument.

(The only thing I can see that this could possibly break is if you are logged in into one wandb account, and tried to pass an API key for a different wandb account. I don't know if this is a concern?)

### DIFF
--- a/openrlhf/trainer/dpo_trainer.py
+++ b/openrlhf/trainer/dpo_trainer.py
@@ -65,7 +65,8 @@ class DPOTrainer(ABC):
             import wandb
 
             self._wandb = wandb
-            wandb.login(key=strategy.args.use_wandb)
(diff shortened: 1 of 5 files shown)


hijkzzz:
Great MR

PR #276, opened 2024-04-26
upgrade transformer/deespeed and sync when colocate models

hijkzzz:


hijkzzz:
I tested it, ant this MR works well now.

PR #283, opened 2024-04-29
fix vLLM v0.4.1

hijkzzz:


hijkzzz:
Convergence testing is also required

hijkzzz:
test failed with TP=1

hijkzzz:
TP=1 fixed

PR #298, opened 2024-05-21
Added GPU memory specs and clarifications, fixed typo.

KT313:
Fixed "NVIDIA A800" to "NVIDIA A100".  
Added gpu memory clarification. I assumed you meant 80GB since this is mentioned in Readme.md line 34.


KT313:
actually i didn't know that A800 GPU exists so i re-corrected it, but i added 40GB mem info since it's not a well-known GPU

hijkzzz:
A800 is the custom version of A100 for China.

PR #328, opened 2024-06-19
Update requirements.txt

Atry:
- `transformers_stream_generator` is removed because it's not in use
- `ray` is unlocked

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,12 +10,11 @@ loralib
 optimum
 packaging
 peft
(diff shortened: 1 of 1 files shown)


catqaq:
ray[default]>=2.12.0

double check please~ @wuxibin89

hijkzzz:
ray[default]==2.12.0 was set to avoid version mismatch

Atry:
Ray 2.24.0 is compatible according to my test

PR #334, opened 2024-06-27
train_rm apply custom tokenizer chat template

mickelliu:
I have a use case where I need to train RM from the base model, it would be helpful to be able to add a chat template here too. Copy paste from train_sft.py.


armsp:
@mickelliu I also have the same use case - train RM with base models :) Would you be so kind to tell me how to actually use a chat template for Base models? Like, how should I pass the argument and should it be `--input_template`, or `--apply_chat_template`, or `--tokenizer_chat_template` or a combination of these?

PR #341, opened 2024-07-01
support remote rm and ref model api for ppo

catqaq:
Simplify RLHF! Use a remote Reward Model (RM) and Reference model, break free from the constraints of the RM and reference model, and focus solely on training the actor and critic models. This significantly reduces the GPU memory requirements and further unleashes the potential of OpenRLHF!

Main modification points:
- remote rm/ref api: openrlhf/utils/utils_for_api.py
- ppo related:
  - examples/train_ppo.py
  - openrlhf/models/model.py
  - openrlhf/trainer/ppo_trainer.py
  - openrlhf/trainer/ppo_utils/experience_maker.py
  - openrlhf/utils/deepspeed.py
-  sh script
- others：
  - avoid conflicts: openrlhf/utils/logging.py -> openrlhf/utils/logging_utils.py
  - openrlhf/trainer/ray/vllm_worker_wrap.py

Note, this is a simple implementation I hastily put together tonight. If there are no issues and it doesn't affect other modules, it can be merged first; further optimization can be done later when time permits, as I may not have more time available in the near term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ tokenizer.apply_chat_template(dataset[0]["input_key"], tokenize=False)
 ```
 > [!NOTE]
 > By default, we use `train` and `test` as splits to distinguish training and testing datasets from Huggingface.
-> The ``JSON key`` options depends on the specific datasets. See [Reward Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/895e8089dc0b1db230316207ca702d5133ae18fd/openrlhf/datasets/reward_dataset.py#L10) and [SFT Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/895e8089dc0b1db230316207ca702d5133ae18fd/openrlhf/datasets/sft_dataset.py#L9)
+> The ``JSON key`` options depends on the specific datasets. See [Reward Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/9ab08aace2c9af7dfa7d8790380d400902deef00/openrlhf/datasets/reward_dataset.py#L10) and [SFT Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/9ab08aace2c9af7dfa7d8790380d400902deef00/openrlhf/datasets/sft_dataset.py#L9)
 
 ### Supervised Fine-tuning
 
--- a/README_zh.md
+++ b/README_zh.md
@@ -138,8 +138,8 @@ tokenizer.apply_chat_template(dataset[0]["input_key"], tokenize=False)
 "<s>[INST] Hello, how are you? [/INST]I'm doing great. How can I help you today?</s> [INST] I'd like to show off how chat templating works! [/INST]"
 ```
 > [!NOTE]
->默认情况下我们使用 `train` 和 `test` 作为 split 区分 Huggingface 的训练/测试数据
->`JSON key` 选项取决于具体的数据集。请参阅 [Reward Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/895e8089dc0b1db230316207ca702d5133ae18fd/openrlhf/datasets/reward_dataset.py#L10) 和 [SFT Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/895e8089dc0b1db230316207ca702d5133ae18fd/openrlhf/datasets/sft_dataset.py#L9)
+>默认情况下我们使用 `train` 和 `test` 作为 split 区分 Huggingface 的训练/测试数据。
+>`JSON key` 选项取决于具体的数据集。请参阅 [Reward Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/9ab08aace2c9af7dfa7d8790380d400902deef00/openrlhf/datasets/reward_dataset.py#L10) 和 [SFT Dataset](https://github.com/OpenLLMAI/OpenRLHF/blob/9ab08aace2c9af7dfa7d8790380d400902deef00/openrlhf/datasets/sft_dataset.py#L9)。
 
 
 ### Supervised Fine-tuning


catqaq:
I have just passed the test on version 0.2.6, but I am still testing on the latest version.

hijkzzz:
I can understand the significance of remote RM (Reward Model), but it seems that a remote Reference Model doesn't make much sense. Additionally, there could be issues with implementing the Reference Model, because if the tokenizers and base models on both sides are not aligned, it could cause problems with the calculation of advantages.

openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -124,13 +131,34 @@ def make_experience(self, prompts: Union[str, List[str]], **generate_kwargs) ->
         action_log_probs = self.actor(sequences, num_actions, attention_mask)
 
         # init log probs
-        base_action_log_probs = self.initial_model(sequences, num_actions, attention_mask)
+        if self.remote_ref_url is not None:
+            # remote ref model
+            # TODO: decoded responses may contain some special tokens：'<|im_start|>','<|im_end|>'
+            responses = self.tokenizer.batch_decode(sequences.cpu(), skip_special_tokens=True)
```
hijkzzz, line 137: `skip_special_tokens=True` will cause a bug because the token can not be aligned.

catqaq:
> I can understand the significance of remote RM (such as for Nemotron 340B RM), but it seems that a remote Reference Model doesn't make much sense. Additionally, there could be issues with implementing the Reference Model, because if the tokenizers and base models on both sides are not aligned, it could cause errors with the calculation of the advantages/loss function.

Yes, the use of a remote reference model needs to be approached with caution, typically requiring the same tokenizer as the actor model. During testing, I used the same model. However, in fact, there are some interesting things that can be done here, but they are not currently a high priority. Therefore, for the time being, I have considered it as an optional feature, as there are indeed some risks associated with direct use due to tokenizer issues and the consistency of encoding and decoding.

openrlhf/trainer/ppo_utils/experience_maker.py
catqaq: > `skip_special_tokens=True` for the Reference Model will cause a bug because the tokens may not be aligned between the actor and reference model.

Even if `skip_special_tokens=True` is specified, the decoding results still include special tokens, such as '<|im_start|>', '<|im_end|>'. This seems to be related to the chat template and the encoding-decoding behavior of the tokenizer. I don't have the energy to focus on the details for now. Under the assumption that the tokenizers are the same, are there any good ways to ensure the reversibility of encoding and decoding? Or can we just use `input_ids` directly, as long as the service also supports it?

hijkzzz:
Update:

Use Ray to use low-compute machines

In Ray, you can control task scheduling by specifying the node's IP address. Ray allows you to specify resource constraints when submitting tasks, and you can use custom resource labels to help the scheduler select the appropriate node. Here is a basic example demonstrating how to use Ray's resource labels and IP addresses to specify nodes:

Start Ray on the nodes:

When starting Ray on each node, you can specify custom resource labels. For example:

On machine A (with small memory):

```bash
ray start --node-ip-address=<IP of machine A> --resources '{"small_memory": 1}'
```
On machine B (with large memory):

```bash
ray start --node-ip-address=<IP of machine B> --resources '{"large_memory": 1}'
```
Specify resource requirements in your script:

When submitting tasks, you can specify the resources required for the task. For example:

```python
import ray

ray.init(address='auto')  # Connect to the Ray cluster

@ray.remote(resources={"small_memory": 1})
def task1():
    # Task suitable for small memory
    pass

@ray.remote(resources={"large_memory": 1})
def task2():
    # Task suitable for large memory
    pass

# Submit tasks
result1 = task1.remote()
result2 = task2.remote()
```

In this example, task1 will be scheduled on a node with the small_memory resource (i.e., machine A), and task2 will be scheduled on a node with the large_memory resource (i.e., machine B).

This method allows you to achieve node scheduling through resource labels without directly using IP addresses. If you really need to specify the node IP directly, you could consider combining Ray's Node API, but this approach is generally not recommended as it breaks the abstraction and flexibility of task scheduling.

zaemyung:
@hijkzzz Could you provide more example codes for using Ray for the cases where a reward model is served by one GPU on the same single node? More specifically, where should we modify to support API-based calls?

hijkzzz:
> @hijkzzz Could you provide more example codes for using Ray for the cases where a reward model is served by one GPU on the same single node? More specifically, where should we modify to support API-based calls?

I'm not sure I get what you mean,

Tou can modify the resources-related code in [train_ppo_ray.py](https://github.com/OpenLLMAI/OpenRLHF/blob/main/examples/train_ppo_ray.py). For example, we want to deploy the Reference Model on a V100 x8 node:

```
ray start --node-ip-address=<IP of machine A> --resources '{"v100": 8}'

# Modify
ref_model = PPORayActorGroup(
        args.ref_num_nodes,
        args.ref_num_gpus_per_node,
        ReferenceModelRayActor,
        pg=pg,
        num_gpus_per_actor=0.25 if pg else 1,
    )

# To
ref_model = PPORayActorGroup(
        args.ref_num_nodes,
        args.ref_num_gpus_per_node,
        ReferenceModelRayActor,
        pg=pg,
        num_gpus_per_actor=1,
        resources={"v100": 1}
        num_resources_per_node=8,
    )
```

zaemyung:
@hijkzzz I see. Thanks for the pointer! 🙏🏻
I can see that essentially the reward model (nn.Module) will be deployed and later called for its [`forward` method](https://github.com/OpenLLMAI/OpenRLHF/blob/main/openrlhf/trainer/ppo_utils/experience_maker.py#L263).

I think this is a different question, but my reward model is essentially a prompt-based one, wrapping an existing LLM (without the value head) with custom class for prompting and constrained generation. In this case, what would be an appropriate way of interaction?

hijkzzz:
> @hijkzzz I see. Thanks for the pointer! 🙏🏻 I can see that essentially the reward model (nn.Module) will be deployed and later called for its [`forward` method](https://github.com/OpenLLMAI/OpenRLHF/blob/main/openrlhf/trainer/ppo_utils/experience_maker.py#L263).
> 
> I think this is a different question, but my reward model is essentially a prompt-based one, wrapping an existing LLM (without the value head) with a custom class for prompting and constrained generation (more specifically I'm using Qwen loaded up using Ollama). In this case, what would be an appropriate way of interaction?

You can load it with vLLM, but you need to modify some code, refer to the vLLM Engine https://github.com/OpenLLMAI/OpenRLHF/blob/main/openrlhf/trainer/ray/vllm_engine.py.

@catqaq The remote reward model makes a lot of sense in this scenario, and it is recommended that this feature be implemented first and included in Ray PPO.

catqaq (commented):
fix conflicts

PR #363, opened 2024-07-16
update "reward_dataset" to properly handle "prompt_key" when "apply_chat_template==True"

Nickydusk:
When `prompt_key` is specified, use it to select the appropriate prompt instead of the current behavior that set `prompt=""`, which will ignore all prompts when `apply_chat_template==True`

### DIFF
--- a/openrlhf/datasets/reward_dataset.py
+++ b/openrlhf/datasets/reward_dataset.py
@@ -17,14 +17,19 @@ def preprocess_data(
     is_dpo=False,
 ) -> str:
     if apply_chat_template:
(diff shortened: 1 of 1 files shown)


hijkzzz:
`prompt_key` it not required when using `apply_chat_template`

Nickydusk:
I am not very certain with the standard usage of the tokenizer, but in my use case, I have processed the "prompt, chosen, and rejected" all into [{role: , content: }] format. Therefore, I need to handle it this way. Is there a problem with my usage?

PR #375, opened 2024-07-21
Improve desc of tqdm

tongyx361:
Add `desc` for `tqdm` without it, providing a better perception of the job process.


catqaq (commented):
thx, it's clearer

PR #422, opened 2024-08-16
add 'num_return_sequences' feature in actor

Nickydusk:
The interface `num_return_sequences` is reserved for algorithms that require multiple samplings for a single prompt, such as [GRPO](https://github.com/deepseek-ai/DeepSeek-Math), which I'm currently developing. It should have no effect on current algorithms (e.g. PPO) since 'num_return_sequences' will use default value '1' when not specified.

### DIFF
--- a/openrlhf/models/actor.py
+++ b/openrlhf/models/actor.py
@@ -132,6 +132,7 @@ class Actor(nn.Module):
             "eos_token_id": kwargs.get("eos_token_id"),
             "pad_token_id": kwargs.get("pad_token_id"),
             "min_new_tokens": kwargs.get("min_new_tokens", 1),
+            "num_return_sequences": kwargs.get("num_return_sequences", 1),
         }
 
         if kwargs.get("max_new_tokens", None):


hijkzzz:
Same as https://github.com/OpenRLHF/OpenRLHF/blob/e1205adaf163e28ed58400ed84618bc01c3c4c37/openrlhf/cli/train_ppo_ray.py#L263

Nickydusk:
I may have overlooked some parts, but it seems to me that the `--n_samples_per_prompt` parameter was only specified and not actually utilized. I thought it should appear at these places.

https://github.com/OpenRLHF/OpenRLHF/blob/e1205adaf163e28ed58400ed84618bc01c3c4c37/openrlhf/models/actor.py#L122-L135

https://github.com/OpenRLHF/OpenRLHF/blob/e1205adaf163e28ed58400ed84618bc01c3c4c37/openrlhf/trainer/ppo_utils/experience_maker.py#L365-L372

hijkzzz:
It was used in prompt_dataset

Nickydusk:
I got it. But I would like to further explore if there is a potential reason for such an indirect design? Using the parameters during generation (e.g., `--num_return_sequences` for huggingface transformers and `--n` for vllm) could be a more logically direct and memory-efficient approach.

PR #449, opened 2024-09-25
Support packing_samples for ppo with ray

zhuzilin:
This is an attempt to add packing_samples to train_ppo_ray.

The main ideas of this PR are:
-  Store data in `Experience` as a list of tensor instead of a padded square shape tensor, and concat them after the dataloader/shuffling;
- Use a `list[int]` of `num_actions` to record the length of the responses.

Currently, the `remote_rm_url`, no vllm and `pretrain_data` are not supported. It should not hard to gradually support them after the changes in this PR are settled.

And ongoing testing of the PR is recorded in: https://wandb.ai/zhuzilin/openrlhf_train_ppo_packing/workspace

With the script:
```bash
   --runtime-env-json='{"env_vars": {"working_dir": "/openrlhf"}}' \
   --ref_num_nodes 1 \
   --ref_num_gpus_per_node 2 \
   --reward_num_nodes 1 \
   --reward_num_gpus_per_node 2 \
   --critic_num_nodes 1 \
   --critic_num_gpus_per_node 2 \
   --actor_num_nodes 1 \
   --actor_num_gpus_per_node 2 \
   --vllm_num_engines 2 \
   --colocate_critic_reward \
   --colocate_actor_ref \
   --reward_pretrain OpenRLHF/Llama-3-8b-rm-mixture \
   --save_path /openrlhf/examples/checkpoint/llama3-8b-rlhf \
   --micro_train_batch_size 4 \
   --train_batch_size 128 \
   --micro_rollout_batch_size 16 \
   --rollout_batch_size 1024 \
   --max_samples 100000 \
   --max_epochs 1 \
   --prompt_max_len 1024 \
   --generate_max_len 1024 \
   --zero_stage 3 \
   --adam_offload \
   --bf16 \
   --actor_learning_rate 5e-7 \
   --critic_learning_rate 9e-6 \
   --init_kl_coef 0.01 \
   --apply_chat_template \
   --normalize_reward \
   --flash_attn \
   --gradient_checkpointing \
   --packing_samples \
   --enforce_eager
```

P.S. The `--enfore_eager` is a common config to disable cuda graph in vllm to save memory.


catqaq:
thx! good job.

1.Do you have more comprehensive tests? Are the final losses or other metrics comparable? What are the differences between these two experiments?
image
2.--packing_samples with remote rm url, this feature depends on the design of the RM API, and we will gradually support it in the future.
3. enfore_eager will harm performance?
According to [the documentation](https://docs.vllm.ai/en/latest/models/engine_args.html), it disables the construction of CUDA graph in Pytorch. **This may harm performance** but reduces the memory requirement (of maintaining the CUDA graph)
https://github.com/vllm-project/vllm/issues/4449

zhuzilin:
@catqaq 
> Do you have more comprehensive tests? Are the final losses or other metrics comparable? What are the differences between these two experiments?

I think the reason of the difference between policy loss with and without `--packing_samples` is that the not packed version will do "left padding"
https://github.com/OpenRLHF/OpenRLHF/blob/4c902a2a35e20072ba949f1f09efe316809c363e/openrlhf/trainer/ppo_utils/replay_buffer.py#L105

when making batches, which means the inputs of the policy are `[pad] [pad] ... [pad] prompt response`. This is not aligned to the pretrain / sft setting of the policy model and thus causing a quite shaking loss. This will be avoided with the packing version.

> enfore_eager will harm performance?

This is not a manditory config for using `--packing_samples` just a common config of vllm to turn off cuda graph, trading a little performance with several GB of memory.

PR #451, opened 2024-09-29
add tensorboard for local use

catqaq:
A very simple logging feature, but many users have been waiting for a long time, haha. Everyone can help check if there are any omissions, thank you.


hijkzzz:
It might be better for use_tensorboard to set the logs path, otherwise users won't know where the logs are.

catqaq:
> It might be better for use_tensorboard to set the logs path? otherwise users won't know where the logs are.

yep

PR #457, opened 2024-10-07
fixed the missing value_head_prefix

ChenmienTan:


hijkzzz:
please also fix the `with deepspeed.zero.GatheredParameters([model.value_head.weight], modifier_rank=0):`
thanks

PR #458, opened 2024-10-08
[BUG] fix _max_steps not initialized bug

BeingGod:
When I do RLHF task it occured an attribute not found bug due to `_max_steps` may not be initialized before using it.
image

reproduce:
```bash
bash OpenRLHF/examples/scripts/train_ppo_llama_ray.sh
```


BeingGod:
PTAL @xiaoxigua999

hijkzzz (approved):

PR #463, opened 2024-10-15
Move the n_samples_per_prompt into replay buffer

zhuzilin:
This PR is trying to change the behavior of `n_samples_per_prompt`. The old behavior is similar to increase epoch, while the new behavior is repetitively sample the same prompt within the same reply buffer.

This is also an preparation for normalizing the reward as GRPO and separate the generation batch size and `rollout_batch_size` for ref model and critic model.

This is a testing wandb for loading checkpoint:
image

The unmatch part is due to the randomness of vllm sampling.

### DIFF
--- a/openrlhf/datasets/prompts_dataset.py
+++ b/openrlhf/datasets/prompts_dataset.py
@@ -5,7 +5,10 @@ from .utils import exist_and_not_none
 
 def preprocess_data(data, input_template=None, input_key="input", apply_chat_template=None) -> str:
     if apply_chat_template:
-        prompt = apply_chat_template(data[input_key], tokenize=False, add_generation_prompt=True)
+        chat = data[input_key]
+        if isinstance(chat, str):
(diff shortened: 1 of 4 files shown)


openrlhf/datasets/prompts_dataset.py
```diff
@@ -5,7 +5,10 @@
 
 def preprocess_data(data, input_template=None, input_key="input", apply_chat_template=None) -> str:
     if apply_chat_template:
-        prompt = apply_chat_template(data[input_key], tokenize=False, add_generation_prompt=True)
+        chat = data[input_key]
+        if isinstance(chat, str):
+            chat = [{"role": "user", "content": chat}]
+        prompt = apply_chat_template(chat, tokenize=False, add_generation_prompt=True)
```
zhuzilin, line 11: This is irrelevant with the main purpose of the PR, but will allow more dataset be used as prompt dataset.

PR #466, opened 2024-10-18
Separate the rollout generation and advantage calculation

zhuzilin:
This PR is trying to turn `make_experience` into 3 steps:

1. Generate the responses from prompts and get the reward score from reward models.
2. Process all the reward values of a whole rollout, e.g. normalize them as in the the GRPO paper, or filter the samples as in PF-PPO.
3. Calculate the advantages and returns with the processed samples.

Here is the comparison with `--packing_samples` (pink line is this PR, blue is the old code):
image

Here is the comparison without `--packing_samples`:
image

---

The above 2 experiments are run with shuffle disabled (both the prompt dataloader and the replay buffer dataloader are set with `shuffle=False`). That's the reason of the anormalies in the training process.

With `shuffle=True`, I rerun the experiments with `--packing_samples`, and the results is (dark red line is this PR with shuffle, grey line is the old code with shuffle):

image

### DIFF
--- a/openrlhf/models/utils.py
+++ b/openrlhf/models/utils.py
@@ -30,20 +30,17 @@ def compute_approx_kl(
 def compute_reward(
     r: Union[torch.Tensor, float],
     kl_coef: float,
-    log_probs: torch.Tensor,
-    log_probs_base: torch.Tensor,
+    kl: Union[torch.Tensor, list[torch.Tensor]],
     action_mask: Optional[torch.Tensor] = None,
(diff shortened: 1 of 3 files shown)


zhuzilin:
@hijkzzz I think the reason for the anomalies in the kl loss/reward is that I set `shuffle=False` to make the experiment more deterministic.... And I've run a new set of experiments with `shuffle=True`. Those look fine and I've added the figure to the PR description comment~

hijkzzz (approved):

PR #473, opened 2024-10-22
 remove unnecessary softmax in prm loss

catqaq:
Two consecutive softmax operations can cause some negative effects, while a normal single softmax would perform better (and won't lead to any crashes; in fact, there is no fundamental difference between one softmax and two softmax operations). As for the specific impact of using two softmax operations, I will explain it in detail in a future blog. Essentially, it functions in a way similar to a temperature coefficient.


catqaq:
some baseline：
image

zhuzilin (approved):

PR #474, opened 2024-10-23
Upload experience_maker perf status

zhuzilin:
This PR tries to upload some performance stats so that we could monitor the bottleneck of the train process easier.
A better way to do that would be writing a perform collecting class... but that may be over-engineering at the moment.

### DIFF
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -113,6 +113,7 @@ class NaiveExperienceMaker(ABC):
         self.kl_ctl = kl_controller
         self.strategy = strategy
         self.reward_fn = reward_fn
+        self.perf_stats = None
 
     # tokenizer
     def tokenize_fn(self, texts, max_length, padding=True, device=None):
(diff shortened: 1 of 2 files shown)


openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -438,10 +445,9 @@ def make_experience(self, prompts: Union[str, List[str]], **generate_kwargs) ->
         }
 
         if self.strategy.args.perf:
-            batch_size = 1 if isinstance(prompts, str) else len(prompts)
-            info["generate_time"] = torch.full((batch_size,), generate_time, device=device)
-            info["actor_time"] = torch.full((batch_size,), actor_time, device=device)
-            info["wait_time"] = torch.full((batch_size,), wait_time, device=device)
+            self.perf_stats["generate_time"] += generate_time
```
hijkzzz, line 448: Why not record the average value?

zhuzilin: hmm... I was think of uploading both the total training time for one rollout and the total time for generating the rollout, so that we can check how much time does the rollout generation take in the ppo training. And with that information, user can see if the bottleneck is in rollout generation or in training, and can adjust the gpu allocation for the bottleneck accordingly.

If we record the average value here, it will be hard to provide the above information.

PR #476, opened 2024-10-24
Support non negative kl divergence approximation

zhuzilin:
This PR is trying to add a non-negative variant of kl divergence approximation. The formula is (from DeepSeekMath https://arxiv.org/abs/2402.03300):
image

It actually correspond to the k3 method in http://joschu.net/blog/kl-approx.html, and the old method is k1. Both of them are unbiased estimators.


catqaq (commented):
can we use use_kl_estimator_k3 -> kl_estimator_type?

PR #477, opened 2024-10-24
Merge Ring Attention into SFT Trainer

ZetangForward:
I add the ring attention to the SFT Trainer.

The openrlhf/datasets/sft_dataset.py file is modified based on the https://github.com/OpenRLHF/OpenRLHF/blob/2a09aeedc10b0f79930673114f28bd48c4e2c5f3/openrlhf/datasets/reward_dataset.py

The `def fit(...)` function is modified based on the code snippet below:
https://github.com/OpenRLHF/OpenRLHF/blob/2a09aeedc10b0f79930673114f28bd48c4e2c5f3/openrlhf/trainer/dpo_trainer.py#L406

Performance & debug log：

Vanilla Loss:

img_v3_02fu_5ce638da-b09b-4d07-9c5d-28fc20b2882g

Ring attention Loss

img_v3_02fv_26f85c0a-6eee-45ad-b23a-26a1992b42cg

Just ignore the order of the samples, the loss of the two attention points is the same~


zhuzilin:
Thank you for the PR. Could you run do a comparison run with and without the ring attention feature and shared the wandb?

ZetangForward:
@zhuzilin hi, my wandb is here, plz check out: https://wandb.ai/zecheng-tang/debug_openrlhf_train_sft/workspace?nw=nwuserzechengtang

Besides, I noticed a very important thing: **the loss accumulation logic seems strange in the raw code**:
https://github.com/OpenRLHF/OpenRLHF/blame/5f52e354829e402561313a8c56e0a5c071a3ae65/openrlhf/trainer/sft_trainer.py#L164

In fact, the normal logic for gradient accumulation should 
1. involve the loss being continuously accumulated through `self.strategy.backward`, and then updating the optimizer when `step % self.strategy.accumulated_gradients` is reached; 
2. At the same time, the gradient (loss) for each step should be divided by `accumulated_gradients` rather than directly propagating backward as in the original code

Here's a code snippet that illustrates the correct way to handle gradient accumulation:
```python
# Assuming self.strategy.accumulated_gradients is the number of steps to accumulate gradients
# and self.optimizer is the optimizer instance.

for step, batch in enumerate(train_loader):
    optimizer.zero_grad()
    outputs = model(batch)
    loss = criterion(outputs, targets)
    loss = loss / self.strategy.accumulated_gradients  # Normalize the loss
    self.strategy.backward(loss)  # Accumulate gradients
    if (step + 1) % self.strategy.accumulated_gradients == 0:
        optimizer.step()  # Update the optimizer
        optimizer.zero_grad()  # Reset gradients for the next accumulation cycle
```

When I tested, I found that the original code would cause a significant deviation when ring attention is turned on or off. The reason is that _when ring attention is turned on, there will inevitably be gradient accumulation (because a batch of data is not as much as vanilla attention), but using the original code would result in "fake gradient accumulation."

In my modified code, the result is the accumulation of gradients, which is why the vanilla loss and ring attention loss curves in wandb basically overlap.



This code snippet ensures that the gradients are accumulated over multiple steps and then normalized before updating the optimizer. This approach should help in reducing the discrepancy you observed when using ring attention.

I'm not sure if my understanding is correct, or if the DeepSpeed within the OpenRLHF library automatically handles the above operations (but upon reviewing the code, it seems like the logic mentioned above is not present). If there are any mistakes, please correct me.

Thank you.

@zhuzilin @tsaoyu

openrlhf/utils/utils.py
```diff
@@ -90,6 +91,12 @@ def blending_datasets(
             train_data = data[train_split].select(range(min(max_count, len(data[train_split]))))
         else:
             train_data = data.select(range(min(max_count, len(data))))
+        
+        # The residual data causes gradient accumulation issues in the last few steps, 
+        # here we automatically discard the extra samples at the end.
+        total_size = len(train_data)
+        keep_size = (total_size // train_batch_size) * train_batch_size
+        train_data = train_data.select(range(keep_size))
```
zhuzilin, line 99: I'm a little confused with the issue here, as the `train_dataloader` has already set with `drop_last=True`, which should serve the same function as the code added here.

openrlhf/datasets/sft_dataset.py
```diff
@@ -91,8 +95,7 @@ def process_data(self, data):
                 add_special_tokens=False,
             )
             prompt_ids_len = prompt_token["attention_mask"].int().sum().item()
-
-            # filter the sample whose length is greater than max_length (2 for answer length)
+            # print(prompt_ids_len)
```
zhuzilin, line 98: please recover the comment here.

openrlhf/datasets/sft_dataset.py
```diff
@@ -81,6 +84,7 @@ def process_data(self, data):
             self.output_key,
             apply_chat_template=None if self.pretrain_mode else self.apply_chat_template,
         )
+
```
zhuzilin, line 87: please remove the newline added here to make the pr clean.

openrlhf/datasets/sft_dataset.py
```diff
@@ -1,11 +1,9 @@
 from typing import Callable
-
 import torch
 from torch.utils.data import Dataset
 from tqdm import tqdm
-
 from .utils import zero_pad_sequences
-
+from torch.nn import functional as F
```
zhuzilin, line 6: please do `pre-commit install` before the commit, which will do auto formatting before commit. Also please recover the removed line here.

openrlhf/cli/train_sft.py
```diff
@@ -52,6 +52,7 @@ def train(args):
         max_count=args.max_samples,
         train_split=args.train_split,
         eval_split=args.eval_split,
+        train_batch_size=args.train_batch_size,
```
zhuzilin, line 55: This doesn't seem to be necessary, as the `blending_datasets` only serve as a function to blend datasets, and it should not require a `train_batch_size`.

openrlhf/datasets/sft_dataset.py
```diff
@@ -64,7 +64,10 @@ def __init__(
 
         # Parallel loading datasets
         processed_dataset = dataset.map(
-            self.process_data, remove_columns=dataset.column_names, num_proc=num_processors
+            self.process_data, 
+            remove_columns=dataset.column_names, 
+            num_proc=num_processors, 
+            keep_in_memory=False,
```
zhuzilin, line 70: please remove the `keep_in_memory=False` which is not in the scope of this PR. If you hope to add this config, it will be nice to propose another PR that apply this change to all the `dataset.map`s.

openrlhf/datasets/sft_dataset.py
```diff
@@ -124,15 +127,14 @@ def __getitem__(self, idx):
...
-
         if not self.pretrain_mode:
             # to avoid EOS_token truncation
             input_token["input_ids"][0][-1] = self.tokenizer.eos_token_id
             input_token["attention_mask"][0][-1] = True
         info = {"input": prompt, "output": response, "input_length": input_token["attention_mask"].int().sum().item()}
-
         return prompt_ids_len, input_token["input_ids"], input_token["attention_mask"], info
 
+
```
zhuzilin, line 137: same as above.

openrlhf/datasets/sft_dataset.py
```diff
@@ -148,23 +150,29 @@ def collate_fn(self, item_list):
...
-        return prompt_ids_lens, input_ids, attention_masks, infos
+        return prompt_ids_lens, input_ids, attention_masks, None, infos
+
 
     def packing_collate_fn(self, item_list):
-        packed_input_ids = []
-        packed_attention_masks = []
-        prompt_ids_lens = []
+        packed_input_ids, packed_attention_masks = [], []
+        prompt_ids_lens, packed_seq_lens = [], []
```
zhuzilin, line 158: please keep the origin code and add a newline for `packed_seq_lens`
```suggestion
        packed_seq_lens = []
```

openrlhf/trainer/sft_trainer.py
```diff
@@ -158,28 +150,84 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
...
+                        "lr": self.scheduler.get_last_lr()[0],
+                    }
+                    if self.aux_loss:
+                        logs_dict["aux_loss"] = accumulated_aux_loss
+                    logs_dict = self.strategy.all_reduce(logs_dict)
+                    step_bar.set_postfix(logs_dict)
+
+                    self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler) # update model parameters
+                    self.optimizer.zero_grad() # clear gradients
+                    self.scheduler.step() # update learning rate
```
zhuzilin, line 229: The scheduler and optimizer will be updated by deepspeed, please don't do the manual update.

zhuzilin (changes requested):

openrlhf/trainer/sft_trainer.py
```diff
@@ -127,28 +127,20 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
             # train
             self.model.train()
             loss_mean = 0
-            for prompts_id_lens, inputs, attention_masks, infos in self.train_dataloader:
+            accumulated_loss = 0
+            accumulated_gpt_loss = 0
+            accumulated_aux_loss = 0
```
zhuzilin, line 132: these variables should be removed?

catqaq:
@ZetangForward thx.
For the ring attention for the SFT, Your implementation is essentially aligned with the SFT Vanilla implementation.
-> in a ring_attn_group, use total token loss/ total token num.
For gradient accumulation, DeepSpeed will handle it automatically. For the current progress, it's recommended that you temporarily remove the manual gradient accumulation implementation as @zhuzilin mentioned.

In fact, as long as gradient accumulation is used, there is an averaging among samples within the batch. If true token-level averaging is required, it’s necessary to bypass DeepSpeed's default gradient accumulation and manually use 'total token loss / total token num in a batch.' Of course, this is a completely separate matter, involving the differences between batch token average, sample token average, and then sample average, which we will discuss in depth later.

openrlhf/trainer/sft_trainer.py
```diff
@@ -158,28 +150,84 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
...
+                    local_mask = (local_label == self.loss_fn.IGNORE_INDEX)  # Shape: (batch_size, seq_len)
+
+                    # convert -100 in local_label into 0 for `torch.gather` operation
+                    local_label[local_mask] = 0
+                    per_token_logps = torch.gather(local_logits.log_softmax(-1), dim=2, index=local_label.unsqueeze(2)).squeeze(2)
+
+                    per_token_logps = per_token_logps * (~local_mask)
+                    gathered_logps = all_gather(per_token_logps, self.strategy.ring_attn_group)
+
+                    gpt_loss = -torch.sum(gathered_logps) / num_calculate_tokens # compute loss on non-masked tokens
```
catqaq, line 193: hi, num_calculate_tokens seems not defined?

openrlhf/trainer/sft_trainer.py
```diff
@@ -127,28 +127,20 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
             # train
             self.model.train()
             loss_mean = 0
-            for prompts_id_lens, inputs, attention_masks, infos in self.train_dataloader:
+            accumulated_loss = 0
```
catqaq, line 130: as @zhuzilin mentioned, if you don't want to do average in a batch, just remove it. Now, although the accumulated_loss is accumulated, it is not used afterward.

catqaq (changes requested):

openrlhf/utils/utils.py
ZetangForward: I have already rolled back the files. Regarding the modifications to the framework, I will be opening a separate discussion. This will help ensure that our discussions remain focused and that all changes are well-documented and understood by the community.

openrlhf/datasets/sft_dataset.py
ZetangForward: solved

openrlhf/trainer/sft_trainer.py
ZetangForward: This variable is only utilized for logging, as the logs_dict in the original code can only log the loss in the minibatch, not work for accumulated_loss

```
logs_dict = {
                        "gpt_loss": gpt_loss.item(),
                        "loss_mean": loss_mean,
                        "lr": self.scheduler.get_last_lr()[0],
                    }
```

ZetangForward:
FYI @zhuzilin @catqaq 

Apologies for the previous PR, in which I introduced too many changes. 

I have since **rolled back the code to the main branch of OpenRLHF** and reverted the modifications related to loss accumulation, ensuring that the framework remains as close as possible to its original state.

Regarding gradient accumulation, I plan to raise this in **a discussion for further exploration.**

In this round of changes, I have made the necessary adjustments as per the requirements, and the behavior in wandb is consistent with what it was before.

openrlhf/cli/train_sft.py
ZetangForward: already fix

openrlhf/trainer/sft_trainer.py
```diff
@@ -159,11 +148,63 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
...
+                else:
+                    aux_loss = 0
+
                 loss = gpt_loss + aux_loss * self.args.aux_loss_coef
                 self.strategy.backward(loss, self.model, self.optimizer)
                 self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler)
 
+                step_bar.update()
+                torch.cuda.empty_cache()  # manually release GPU cache memory to avoid OOM
+
```
zhuzilin, line 207: the `step_bar.update()` has been updated below. And I suggest not adding `torch.cuda.empty_cache()` here by default because it will harm the performance.

ZetangForward: already fix.

catqaq:
ok, thx
"Regarding gradient accumulation, I plan to raise this in a discussion for further exploration."  as I mentioned above, it's related to token average loss, we can discuss it later.

zhuzilin (approved):

PR #480, opened 2024-10-26
fix packing_samples in NaiveExperienceMaker

zmzhang2000:
Fix packing_samples that triggers AttributeError when using NaiveExperienceMaker


hijkzzz:
we can use `getattr(self, "packing_samples")` ?

zmzhang2000:
> Can we use `not getattr(self, "packing_samples", False)` ?

New request (#484) has been pulled to fix this bug.

Thanks for your suggestion.

PR #511, opened 2024-11-08
Allow arbitrary number of vllm engines

zhuzilin:


hijkzzz:
I've added some annotations, but there's still an issue: theoretically, there shouldn't be empty samples for any vLLM engine. I'll think about how to better enforce this condition.

zhuzilin:
> theoretically, there shouldn't be empty samples for any vLLM engine. I'll think about how to better enforce this condition.

Suppose we have `n` prompts and `m` vllm engines for one actor process, and they have the following relation:

`n = km + b, (0 <= b < m)`

An ideal way to distribute the prompts would be the first `b` engines have `k + 1` prompts while the last `m - b` engines have `k` prompts.

And the current impl is that when `b > 0`, give the first `m - 1` engines `k + 1` prompts and the last engine takes the remaining, which is `km + b - (k + 1)(m - 1) = b - m + k + 1`.

We can change the code into the ideal way, and the reason for the current impl is that it's short. And as both allocation will result in a vllm engine with `k + 1` prompts, so the speed will be similar.

PR #514, opened 2024-11-10
[DRAFT] Add group_norm in train_ppo.py and train_ppo_ray.py

xiaoxigua999:


xiaoxigua999:
Work in progress. There are some bugs here.

hijkzzz:
PPO (Red), GRPO (Brown), Reinforce (Green)
image

openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -332,7 +376,10 @@ def make_experience(self, samples: Samples) -> Experience:
     @torch.no_grad()
     def process_experiences(self, experiences: List[Experience]) -> List[Experience]:
         # TODO: add more methods to process experiences
-        return experiences
+        if self.advantage_estimator in ["group_norm"]:
+            return [sum(experiences)]
```
zhuzilin, line 380: I think there is problem when `--packing_samples` is off, because in that case, tensors in each experiences will be padded into difference length.

hijkzzz:
For Rule-based RM

765795cac2b1cf4ad1e332126e29bc0

Wang895:
> For Rule-based RM
> 
> 765795cac2b1cf4ad1e332126e29bc0


How to Implement PPO Training with Rule-based RM?

PR #518, opened 2024-11-12
Raise warning on faulty input template

cemiu:
Passing `--input_template '\n'` as an arg in Bash/Zsh passes the string '\\n' to Python (2 characters), rather than a new line character. The correct syntax is: `--input_template $'\n'`

This likely leads to confusion (see: it's wrong in the docs & README).

I've made the CLI modules raise an explicit warning & updated the docs.

I'll open a separate PR for the OpenRLHF-Docs repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ deepspeed --module openrlhf.cli.train_sft \
    --dataset Open-Orca/OpenOrca \
    --input_key question \
    --output_key response \
-   --input_template 'User: {}\nAssistant: ' \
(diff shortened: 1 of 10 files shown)


cemiu:
The problem is likely so frequent (and always unintentional), that maybe an explicit assertion would be useful?

I assume close to every training run using the input_template arg is wrong.

zhuzilin (approved):
LGTM! Yeah, the default escaping behavior of bash is quite annoying.

PR #524, opened 2024-11-17
Add support when RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set

HollowMan6:
RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is used for overriding the auto setting of *_VISIBLE_DEVICES for each actor. This commit address the 2 issues when the flag is set:

1. LOCAL_RANK should reflect the actual value instead of "0", to select the correct GPU.
2. vLLM workers should be forced to use ray even if tensor_parallel_size is 1, as "mp" distributed executor backend won't work correctly when *_VISIBLE_DEVICES is not properly set.

Setting RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES flag is neccessary to get OpenRLHF running on AMD GPUs when ray >= 2.10.0 (Referring to https://github.com/vllm-project/vllm/blob/v0.6.4.post1/Dockerfile.rocm#L131-L132)

PS: We also need the following flag to get this working properly:

```
export VLLM_USE_RAY_SPMD_WORKER=1
export VLLM_USE_RAY_COMPILED_DAG=1
```

If without setting RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES, the
error log is as follows:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "OpenRLHF/openrlhf/cli/train_ppo_ray.py", line 396, in <module>
    train(args)
  File "OpenRLHF/openrlhf/cli/train_ppo_ray.py", line 146, in train
    ray.get(refs)
  File "venv/lib/python3.11/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/worker.py", line 2753, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/worker.py", line 904, in get_objects
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(DeferredCudaCallError): mray::RewardModelRayActor.init_model_from_pretrained()(repr=<openrlhf.trainer.ray.launcher.RewardModelRayActor object>)
  File "venv/lib/python3.11/site-packages/torch/cuda/random.py", line 126, in cb
    default_generator = torch.cuda.default_generators[i]
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range

The above exception was the direct cause of the following exception:

ray::RewardModelRayActor.init_model_from_pretrained()[(repr=<openrlhf.trainer.ray.launcher.RewardModelRayActor object>)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 110, in init_model_from_pretrained
    self._setup_distributed(strategy)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 58, in _setup_distributed
    strategy.setup_distributed()
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 87, in setup_distributed
    torch.cuda.set_device(self.args.local_rank)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 478, in set_device
    torch._C._cuda_setDevice(device)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 338, in _lazy_init
    raise DeferredCudaCallError(msg) from e
torch.cuda.DeferredCudaCallError: CUDA call failed lazily at initialization with error: tuple index out of range

CUDA call was originally invoked at:

  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 110, in init_model_from_pretrained
    self._setup_distributed(strategy)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 58, in _setup_distributed
    strategy.setup_distributed()
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 81, in setup_distributed
    self.set_seed(self.seed)
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 78, in set_seed
    torch.cuda.manual_seed_all(seed)
  File "venv/lib/python3.11/site-packages/torch/cuda/random.py", line 129, in manual_seed_all
    _lazy_call(cb, seed_all=True)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 256, in _lazy_call
    _lazy_seed_tracker.queue_seed_all(callable, traceback.format_stack())
```

### DIFF
--- a/openrlhf/trainer/ray/launcher.py
+++ b/openrlhf/trainer/ray/launcher.py
@@ -9,11 +9,11 @@ from ray.util.placement_group import PlacementGroup, placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
 from openrlhf.models import Actor, get_llm_for_sequence_regression
-from openrlhf.utils import DeepspeedStrategy, get_tokenizer
+from openrlhf.utils import DeepspeedStrategy, ray_noset_visible_devices
 
 
 class DistributedTorchRayActor:
-    def __init__(self, world_size, rank, local_rank, master_addr, master_port):
+    def __init__(self, world_size, rank, master_addr, master_port):
         logging.basicConfig(
             format="%(asctime)s %(levelname)-8s %(message)s",
             level=logging.INFO,
@@ -21,17 +21,17 @@ class DistributedTorchRayActor:
         )
         self._world_size = world_size
         self._rank = rank
-        self._local_rank = local_rank
         self._master_addr = master_addr if master_addr else self._get_current_node_ip()
         self._master_port = master_port if master_port else self._get_free_port()
         os.environ["MASTER_ADDR"] = self._master_addr
         os.environ["MASTER_PORT"] = str(self._master_port)
         os.environ["WORLD_SIZE"] = str(self._world_size)
         os.environ["RANK"] = str(self._rank)
-        # NOTE: Ray will automatically set the CUDA_VISIBLE_DEVICES
-        # environment variable for each actor, so always set device to 0
-        # os.environ["LOCAL_RANK"] = str(self._local_rank)
-        os.environ["LOCAL_RANK"] = "0"
+        # NOTE: Ray will automatically set the *_VISIBLE_DEVICES
+        # environment variable for each actor, unless
+        # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set, so
+        # set local rank to 0 when the flag is not applicable.
+        os.environ["LOCAL_RANK"] = str(ray.get_gpu_ids()[0]) if ray_noset_visible_devices() else "0"
 
     @staticmethod
     def _get_current_node_ip():
@@ -197,20 +197,19 @@ class PPORayActorGroup:
                 scheduling_strategy=PlacementGroupSchedulingStrategy(
                     placement_group=pg, placement_group_bundle_index=0
                 ),
-            ).remote(world_size, 0, 0, None, None)
+            ).remote(world_size, 0, None, None)
         else:
             master_actor = self.ray_actor_type.options(
                 num_cpus=num_gpus_per_actor,
                 num_gpus=num_gpus_per_actor,
                 resources=self._resources,
-            ).remote(world_size, 0, 0, None, None)
+            ).remote(world_size, 0, None, None)
         self._actor_handlers = [master_actor]
 
         # Create worker actors
         if world_size > 1:
             master_addr, master_port = ray.get(master_actor.get_master_addr_port.remote())
             for rank in range(1, world_size):
-                local_rank = rank % self._num_gpus_per_node
                 if pg:
                     worker_actor = self.ray_actor_type.options(
                         num_cpus=num_gpus_per_actor,
@@ -220,13 +219,13 @@ class PPORayActorGroup:
                             placement_group=pg,
                             placement_group_bundle_index=rank // self._num_gpus_per_node,
                         ),
-                    ).remote(world_size, rank, local_rank, master_addr, master_port)
+                    ).remote(world_size, rank, master_addr, master_port)
                 else:
                     worker_actor = self.ray_actor_type.options(
                         num_cpus=num_gpus_per_actor,
                         num_gpus=num_gpus_per_actor,
                         resources=self._resources,
-                    ).remote(world_size, rank, local_rank, master_addr, master_port)
+                    ).remote(world_size, rank, master_addr, master_port)
                 self._actor_handlers.append(worker_actor)
 
     def async_init_model_from_pretrained(
--- a/openrlhf/trainer/ray/vllm_engine.py
+++ b/openrlhf/trainer/ray/vllm_engine.py
@@ -5,10 +5,15 @@ import ray
 from ray.util.placement_group import placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
+from openrlhf.utils import ray_noset_visible_devices
+
 from openrlhf.utils.logging_utils import init_logger
 
 logger = init_logger(__name__)
 
+@ray.remote
+def get_all_env_variables():
+    return os.environ
 
 @ray.remote
 class LLMRayActor:
@@ -18,7 +23,7 @@ class LLMRayActor:
         self.__version__ = vllm.__version__
         assert self.__version__ >= "0.4.1", "OpenRLHF only supports vLLM >= 0.4.1"
 
-        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1
+        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1 and not ray_noset_visible_devices()
 
         # See https://github.com/vllm-project/vllm/blob/main/vllm/executor/gpu_executor.py
         if self.use_gpu_executor:
@@ -83,12 +88,17 @@ def create_vllm_engines(
     max_model_len: int,
 ):
     vllm_engines = []
+    # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES will always be set in current context,
+    # So we need to get env variables from ray process to check if it is set.
+    noset_visible_devices = ray_noset_visible_devices(ray.get(get_all_env_variables.remote()))
     for i in range(num_engines):
-        # When tensor_parallel_size=1, vLLM init model in LLMEngine directly, assign 1 GPU for it.
-        num_gpus = int(tensor_parallel_size == 1)
+        # When tensor_parallel_size=1 and RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is not set
+        # (vLLM mp backend will work smoothly only when *_VISIBLE_DEVICES is modified),
+        # vLLM init model in LLMEngine directly, assign 1 GPU for it.
+        num_gpus = int(tensor_parallel_size == 1 and not noset_visible_devices)
         scheduling_strategy = None
 
-        if tensor_parallel_size > 1:
+        if tensor_parallel_size > 1 or noset_visible_devices:
             bundles = [{"GPU": 1, "CPU": 1}] * tensor_parallel_size
             pg = placement_group(bundles)
             ray.get(pg.ready())
--- a/openrlhf/utils/__init__.py
+++ b/openrlhf/utils/__init__.py
@@ -1,3 +1,3 @@
 from .deepspeed import DeepspeedStrategy
 from .processor import get_processor, reward_normalization
-from .utils import blending_datasets, get_strategy, get_tokenizer
+from .utils import blending_datasets, get_strategy, get_tokenizer, ray_noset_visible_devices
--- a/openrlhf/utils/utils.py
+++ b/openrlhf/utils/utils.py
@@ -12,6 +12,27 @@ DEFAULT_BOS_TOKEN = "<s>"
 DEFAULT_UNK_TOKEN = "<unk>"
 
 
+def ray_noset_visible_devices(env_vars=os.environ):
+    # Refer to
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/nvidia_gpu.py#L95-L96
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/amd_gpu.py#L102-L103
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/npu.py#L94-L95
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/hpu.py#L116-L117
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/neuron.py#L108-L109
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/tpu.py#L171-L172
+    # https://github.com/ray-project/ray/blob/161849364a784442cc659fb9780f1a6adee85fce/python/ray/_private/accelerators/intel_gpu.py#L97-L98
+    NOSET_VISIBLE_DEVICES_ENV_VARS_LIST = [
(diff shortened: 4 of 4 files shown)


hijkzzz:
@wuxibin89 helps to review this MR~

openrlhf/trainer/ray/vllm_engine.py
```diff
@@ -84,11 +86,13 @@ def create_vllm_engines(
 ):
     vllm_engines = []
     for i in range(num_engines):
-        # When tensor_parallel_size=1, vLLM init model in LLMEngine directly, assign 1 GPU for it.
-        num_gpus = int(tensor_parallel_size == 1)
+        # When tensor_parallel_size=1 and RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is not set
+        # (vLLM mp backend will work smoothly only when *_VISIBLE_DEVICES is modified),
+        # vLLM init model in LLMEngine directly, assign 1 GPU for it.
+        num_gpus = int(tensor_parallel_size == 1 and not ray_noset_visible_devices())
```
HollowMan6, line 92: Tested with NVIDIA GPUs and found that somehow `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` is set automatically (maybe by ray) when executing `create_vllm_engines`, working on a solution for this...

HollowMan6: Now this is fixed!

openrlhf/trainer/ray/launcher.py
```diff
@@ -28,10 +28,11 @@ def __init__(self, world_size, rank, local_rank, master_addr, master_port):
...
         os.environ["RANK"] = str(self._rank)
-        # NOTE: Ray will automatically set the CUDA_VISIBLE_DEVICES
-        # environment variable for each actor, so always set device to 0
-        # os.environ["LOCAL_RANK"] = str(self._local_rank)
-        os.environ["LOCAL_RANK"] = "0"
+        # NOTE: Ray will automatically set the *_VISIBLE_DEVICES
+        # environment variable for each actor, unless
+        # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set, so
+        # set local rank to 0 when the flag is not applicable.
+        os.environ["LOCAL_RANK"] = str(self._local_rank) if ray_noset_visible_devices() else "0"
```
wuxibin89, line 35: @HollowMan6 If PPORayActorGroup use whole GPUs of a machine, (that is `num_gpus_per_node=8` for NVIDIA A100/H100, or `num_gpus_per_node=16` for NVIDIA L20), then LOCAL_RANK is the correct device index for this DistributedTorchRayActor. 
But when `num_gpus_per_node=4`, then two PPORayActorGroup may be scheduled on same machine by ray, their local_ranks are both `0,1,2,3`, which leads to incorrect use the same set of GPUs.

HollowMan6: Ah, that's true. I only tested the situation when a full node was occupied. I'll check if we can replace the local_rank with [`ray.get_gpu_ids`](https://docs.ray.io/en/latest/ray-core/api/doc/ray.get_gpu_ids.html)

HollowMan6: Replaced with [ray.get_gpu_ids](https://docs.ray.io/en/latest/ray-core/api/doc/ray.get_gpu_ids.html), cleaned up the old local_rank parameter (turned out to be not correct or useful), and tested. Now it works fine.

wuxibin89 (approved):
LGTM

hijkzzz:
@HollowMan6  I have reverted this MR as I met an error:

```
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464] Error executing method init_device. This might cause deadlock in distributed execution.
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464] Traceback (most recent call last):
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]   File "/usr/local/lib/python3.10/dist-packages/vllm/worker/worker_base.py", line 456, in execute_method
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]     return executor(*args, **kwargs)
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]   File "/usr/local/lib/python3.10/dist-packages/vllm/worker/worker.py", line 166, in init_device
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]     torch.cuda.set_device(self.device)
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]   File "/usr/local/lib/python3.10/dist-packages/torch/cuda/__init__.py", line 420, in set_device
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]     torch._C._cuda_setDevice(device)
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]   File "/usr/local/lib/python3.10/dist-packages/torch/cuda/__init__.py", line 314, in _lazy_init
(LLMRayActor pid=61219) ERROR 11-25 08:11:38 worker_base.py:464]     torch._C._cuda_init()
```

with script
```
set -x

   --runtime-env-json='{"working_dir": "/openrlhf"}' \
   --ref_num_nodes 1 \
   --ref_num_gpus_per_node 1 \
   --reward_num_nodes 1 \
   --reward_num_gpus_per_node 1 \
   --critic_num_nodes 1 \
   --critic_num_gpus_per_node 1 \
   --actor_num_nodes 1 \
   --actor_num_gpus_per_node 1 \
   --vllm_num_engines 1 \
   --colocate_critic_reward \
   --colocate_actor_ref \
   --reward_pretrain OpenRLHF/Llama-3-8b-rm-700k \
   --save_path /openrlhf/examples/test_scripts/checkpoint/llama3-8b-rlhf \
   --micro_train_batch_size 16 \
   --train_batch_size 128 \
   --rollout_batch_size 1024 \
   --n_samples_per_prompt 1 \
   --max_epochs 1 \
   --prompt_max_len 1024 \
   --max_samples 100000 \
   --generate_max_len 1024 \
   --zero_stage 3 \
   --bf16 \
   --actor_learning_rate 5e-7 \
   --critic_learning_rate 9e-6 \
   --init_kl_coef 0.01 \
   --apply_chat_template \
   --normalize_reward \
   --adam_offload \
   --gradient_checkpointing \
   --packing_samples \
   --save_steps -1 \
   --ckpt_path /openrlhf/examples/test_scripts/ckpt/llama3-8b-rlhf 

```

HollowMan6:
> @HollowMan6 I have reverted this MR as I met an error:

@hijkzzz Which vLLM version are you using? Did you set the following environment variables together? Or you didn't set the `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`?

```bash
export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1
export VLLM_USE_RAY_SPMD_WORKER=1
export VLLM_USE_RAY_COMPILED_DAG=1
```

I'm using the latest vLLM version (v0.6.4.post1) and tested your script on my side (NVIDIA cluster). Both of the cases work fine for me.

In any case, it will work when you don't set the `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`, as this is a non-intrusive modification. If the vLLM version causes the issue you are facing, I could add a document saying this is tested to work when the vLLM version is over v0.6.4.

hijkzzz:
>Looks like it happened when you didn't set the RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES, investigating the cause...
> Will reland with a new PR once this is fixed.
@HollowMan6 
Thank you!!

HollowMan6:
Relanding PR opened at https://github.com/OpenRLHF/OpenRLHF/pull/536 I also updated the README for the instructions.

It turns out that the culprit is very strange: it's caused by the call of `ray_noset_visible_devices()` in `LLMRayActor`, although the return value is correct and there's no clear clue how everything is related. A workaround for this I come up with is to pass the value of `ray_noset_visible_devices()` through kwargs. So the only code changes I made compared to the reverted one are these:

- https://github.com/OpenRLHF/OpenRLHF/pull/536/files#diff-d7d5aa9db07c0a235af959a88c3fe0c5440e7441250f9b00009735ba8d0837e6R28-R29
- https://github.com/OpenRLHF/OpenRLHF/pull/536/files#diff-d7d5aa9db07c0a235af959a88c3fe0c5440e7441250f9b00009735ba8d0837e6R120

And it turns out we no longer need `VLLM_USE_RAY_SPMD_WORKER` and `VLLM_USE_RAY_COMPILED_DAG`, even if we set the flag, as those are caused by the exact same culprit (this is actually another workaround for the error). It's still not clear to me how those can be related, I doubt it's a bug related to Ray as possibly it's because `ray_noset_visible_devices()` is not a method of the rayActor and they are not at the same file.

cc: @hijkzzz

HollowMan6:
> It turns out that the culprit is very strange: it's caused by the call of `ray_noset_visible_devices()` in `LLMRayActor`.

Finally figured out how everything is related, this is caused by the side effect of
`from openrlhf.utils.deepspeed import DeepspeedStrategy`, as can be confirmed if you directly put the code under the `__init__` methods of `LLMRayActor`, even if you are not using `DeepspeedStrategy` there.

This was undirectly executed when we excute `ray_noset_visible_devices()`, as the codebase has non-empty `__init__.py` files, which makes the debugging very challenging.

Culprit when `ray_noset_visible_devices` was under `openrlhf/utils/utils.py`:
https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/utils/__init__.py#L1

Culprit when `ray_noset_visible_devices` was under `openrlhf/trainer/ray/utils.py`:
https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/trainer/ray/__init__.py#L1

https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/trainer/ray/launcher.py#L12

As a result, I raised the PR to empty all the __init__ files + clean up unused imports, https://github.com/OpenRLHF/OpenRLHF/pull/537, which hopefully can prevent such things from happening again (at least making debugging much easier).

cc: @hijkzzz

HollowMan6:
Finally, the whole story: RuntimeError: No CUDA GPUs are available is caused by CUDA, as it doesn't dynamically support changing `CUDA_VISIBLE_DEVICES`.

Story line: When `import deepspeed` is executed, the `torch.cuda.is_available()` [here](https://github.com/microsoft/DeepSpeed/blob/f743feca033515fdded50a98093da5a48eb41e74/accelerator/real_accelerator.py#L170) automatically gets executed. At this stage, if `tensor_parallel_size` is not 1, we won't assign any GPU to the `LLMRayActor` process:

https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/trainer/ray/vllm_engine.py#L87-L88

So then if we didn't set `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`, `CUDA_VISIBLE_DEVICES` is an empty string, and CUDA is initialized with no GPU available. 

Later, since we also didn't set `VLLM_USE_RAY_SPMD_WORKER`, for one specific worker that is at the same node of `LLMRayActor` process, vLLM will not use the actual ray reserved process, but rather update the `CUDA_VISIBLE_DEVICES` and directly [start the driver worker in `LLMRayActor` process](https://github.com/vllm-project/vllm/blob/6e9ff050c8e83ad6d5e5eab621e83549e35933a1/vllm/executor/ray_gpu_executor.py#L150-L152). This modification will work if any of the CUDA API hasn't been called, but unfortunately it doesn't because `import deepspeed` is executed.

So, to break everything, it's actually not necessary to be deepspeed related; a simple CUDA call, e.g., `torch.cuda.is_available()`, will already break everything.

Another workaround for the case specifically is to set the `PYTORCH_NVML_BASED_CUDA_CHECK` environment variable to 1, which will use NVML based check for `torch.cuda.is_available()` to mitigate the issue, but again, this doesn't prevent the situation if we make other CUDA calls before we call `vllm.LLM` in `LLMRayActor`.

The ultimate fix would be NVIDIA support changing `CUDA_VISIBLE_DEVICES` dynamically, but I'm not sure if there's any place to make them aware of the situation, because CUDA is close-sourced.

A minimum reproducible code for demo purposes:

```python
import os
import torch

os.environ["CUDA_VISIBLE_DEVICES"] = ""
print(f"torch.cuda.is_available(): {torch.cuda.is_available()}")

os.environ["CUDA_VISIBLE_DEVICES"] = "0,1,2,3"
print(f"torch.cuda.is_available(): {torch.cuda.is_available()}")
print(f"torch.cuda.device_count(): {torch.cuda.set_device('cuda:0')}")
```

Result:

```log
torch.cuda.is_available(): False
torch.cuda.is_available(): False
Traceback (most recent call last):
  File "test.py", line 74, in <module>
    print(f"torch.cuda.device_count(): {torch.cuda.set_device('cuda:0')}")
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 478, in set_device
    torch._C._cuda_setDevice(device)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 319, in _lazy_init
    torch._C._cuda_init()
RuntimeError: No CUDA GPUs are available
```

SparkJiao:
Hi, @HollowMan6 , appreciate your contribution.

After browsing the your PRs, may I confirm if using the latest deepspeed and OpenRLHF can solve this problem?

Thank you very much!

HollowMan6:
Hi @SparkJiao ! I'm not sure what you're referring to, but if you mean running on AMD GPUs, it was working fine after the relanding PR is merged, but haven't checked the latest codebase on AMD GPUs yet. The only thing that I have in mind is that, you should still use gloo as the vLLM sync backend as the default nccl one doesn't work on AMD.

PR #534, opened 2024-11-25
Loosen dependency version requirements

fzyzcjy:
Hi thanks for the library! It would be great if the dependency version requirements can be loosen a little. For example, the recently released transformers 4.46.2 has some nice things to have (e.g. this one https://github.com/huggingface/transformers/pull/34511).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,12 +10,12 @@ loralib
 optimum
 packaging
 peft
-ray[default]>=2.12.0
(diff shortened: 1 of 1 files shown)


hijkzzz:
It's dangerous.

fzyzcjy:
Hmm, then what about updating it to latest

hijkzzz:
merged

PR #536, opened 2024-11-25
[Relanding] Add support when RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set

HollowMan6:
Relanding of PR https://github.com/OpenRLHF/OpenRLHF/pull/524

RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is used for overriding the auto setting of *_VISIBLE_DEVICES for each actor. This commit address the 2 issues when the flag is set:

1. LOCAL_RANK should reflect the actual value instead of "0", to select the correct GPU.
2. vLLM workers should be forced to use ray even if tensor_parallel_size is 1, as "mp" distributed executor backend won't work correctly when *_VISIBLE_DEVICES is not properly set.

Setting RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES flag is neccessary to get OpenRLHF running on AMD GPUs when ray >= 2.10.0 (Referring to https://github.com/vllm-project/vllm/blob/v0.6.4.post1/Dockerfile.rocm#L131-L132)

If without setting RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES, the
error log is as follows:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "OpenRLHF/openrlhf/cli/train_ppo_ray.py", line 396, in <module>
    train(args)
  File "OpenRLHF/openrlhf/cli/train_ppo_ray.py", line 146, in train
    ray.get(refs)
  File "venv/lib/python3.11/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/worker.py", line 2753, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/ray/_private/worker.py", line 904, in get_objects
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(DeferredCudaCallError): mray::RewardModelRayActor.init_model_from_pretrained()(repr=<openrlhf.trainer.ray.launcher.RewardModelRayActor object>)
  File "venv/lib/python3.11/site-packages/torch/cuda/random.py", line 126, in cb
    default_generator = torch.cuda.default_generators[i]
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range

The above exception was the direct cause of the following exception:

ray::RewardModelRayActor.init_model_from_pretrained()[(repr=<openrlhf.trainer.ray.launcher.RewardModelRayActor object>)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 110, in init_model_from_pretrained
    self._setup_distributed(strategy)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 58, in _setup_distributed
    strategy.setup_distributed()
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 87, in setup_distributed
    torch.cuda.set_device(self.args.local_rank)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 478, in set_device
    torch._C._cuda_setDevice(device)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 338, in _lazy_init
    raise DeferredCudaCallError(msg) from e
torch.cuda.DeferredCudaCallError: CUDA call failed lazily at initialization with error: tuple index out of range

CUDA call was originally invoked at:

  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 110, in init_model_from_pretrained
    self._setup_distributed(strategy)
  File "OpenRLHF/openrlhf/trainer/ray/launcher.py", line 58, in _setup_distributed
    strategy.setup_distributed()
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 81, in setup_distributed
    self.set_seed(self.seed)
  File "OpenRLHF/openrlhf/utils/deepspeed.py", line 78, in set_seed
    torch.cuda.manual_seed_all(seed)
  File "venv/lib/python3.11/site-packages/torch/cuda/random.py", line 129, in manual_seed_all
    _lazy_call(cb, seed_all=True)
  File "venv/lib/python3.11/site-packages/torch/cuda/__init__.py", line 256, in _lazy_call
    _lazy_seed_tracker.queue_seed_all(callable, traceback.format_stack())
```

### DIFF
--- a/openrlhf/trainer/ray/vllm_engine.py
+++ b/openrlhf/trainer/ray/vllm_engine.py
@@ -5,9 +5,17 @@ import ray
 from ray.util.placement_group import placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
+from openrlhf.trainer.ray.utils import ray_noset_visible_devices
+
 from openrlhf.utils.logging_utils import init_logger
 
 logger = init_logger(__name__)
+
+
+@ray.remote
+def get_all_env_variables():
+    import os
+    return os.environ
 
 
 @ray.remote
@@ -18,7 +26,8 @@ class LLMRayActor:
         self.__version__ = vllm.__version__
         assert self.__version__ >= "0.4.1", "OpenRLHF only supports vLLM >= 0.4.1"
 
-        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1
+        noset_visible_devices = kwargs.pop("noset_visible_devices", False)
+        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1 and not noset_visible_devices
 
         # See https://github.com/vllm-project/vllm/blob/main/vllm/executor/gpu_executor.py
         if self.use_gpu_executor:
@@ -83,12 +92,17 @@ def create_vllm_engines(
     max_model_len: int,
 ):
     vllm_engines = []
+    # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES will always be set in current context,
+    # So we need to get env variables from ray process to check if it is set.
+    noset_visible_devices = ray_noset_visible_devices(ray.get(get_all_env_variables.remote()))
     for i in range(num_engines):
-        # When tensor_parallel_size=1, vLLM init model in LLMEngine directly, assign 1 GPU for it.
-        num_gpus = int(tensor_parallel_size == 1)
+        # When tensor_parallel_size=1 and RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is not set
+        # (vLLM mp backend will work smoothly only when *_VISIBLE_DEVICES is modified),
+        # vLLM init model in LLMEngine directly, assign 1 GPU for it.
+        num_gpus = int(tensor_parallel_size == 1 and not noset_visible_devices)
         scheduling_strategy = None
 
-        if tensor_parallel_size > 1:
+        if tensor_parallel_size > 1 or noset_visible_devices:
             bundles = [{"GPU": 1, "CPU": 1}] * tensor_parallel_size
             pg = placement_group(bundles)
             ray.get(pg.ready())
@@ -104,6 +118,7 @@ def create_vllm_engines(
                 scheduling_strategy=scheduling_strategy,
             ).remote(
                 pretrain,
+                noset_visible_devices=noset_visible_devices,
                 trust_remote_code=True,
                 tensor_parallel_size=tensor_parallel_size,
                 dtype="bfloat16",
--- a/README.md
+++ b/README.md
@@ -320,6 +320,15 @@ ray job submit --address="http://127.0.0.1:8265" \
 > Do not set `--vllm_num_engines` means not using the vLLM engine.
 > You can also use ``setup_commands`` to let Ray automatically deploy the environment, such as `--runtime-env-json='{"setup_commands": ["pip install openrlhf[vllm]"]}'`.
 
+> [!NOTE]
+> If you want to run on AMD GPUs, or for whatever reason you encounter an error related to index out of range when deepspeed sets up the GPU devices, you can try to set the environment variable [`RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES`](openrlhf/trainer/ray/utils.py) as a workaround.
+>   ```bash
+>   # For NVIDIA GPUs:
+>   export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1
+>   # For AMD GPUs:
+>   export RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES=1
+>   ```
+
 The launch scripts and documents for supported algorithms are in [example/scripts](./examples/scripts/) and [Documents - Usage](https://openrlhf.readthedocs.io/en/latest/usage.html)
 
 ## Performance
--- a/README_zh.md
+++ b/README_zh.md
@@ -328,6 +328,15 @@ ray job submit --address="http://127.0.0.1:8265" \
 > 不设置 `--vllm_num_engines` 则是不使用 vLLM engine。
 > 您也可以通过 ``setup_commands`` 让 Ray 自动初始化环境, 比如 `--runtime-env-json='{"setup_commands": ["pip install openrlhf[vllm]"]}'`
 
+> [!NOTE]
+> 如果您想在 AMD 显卡上运行，或者由于某种原因，在 deepspeed 设置显卡设备时遇到与索引超出范围相关的错误，您可以尝试设置环境变量 [`RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES`](openrlhf/trainer/ray/utils.py)。
+> ```bash
+> # 对于 NVIDIA 显卡:
+> export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1
+> # 对于 AMD 显卡:
+> export RAY_EXPERIMENTAL_NOSET_ROCR_VISIBLE_DEVICES=1
+> ```
+
 所有支持算法的启动脚本和文档在 [example/scripts](./examples/scripts/) 和 [Documents - Usage](https://openrlhf.readthedocs.io/en/latest/usage.html)
 
 
--- a/openrlhf/trainer/ray/launcher.py
+++ b/openrlhf/trainer/ray/launcher.py
@@ -11,9 +11,10 @@ from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 from openrlhf.models import Actor, get_llm_for_sequence_regression
 from openrlhf.utils import DeepspeedStrategy, get_tokenizer
 
+from openrlhf.trainer.ray.utils import ray_noset_visible_devices
 
 class DistributedTorchRayActor:
-    def __init__(self, world_size, rank, local_rank, master_addr, master_port):
+    def __init__(self, world_size, rank, master_addr, master_port):
         logging.basicConfig(
             format="%(asctime)s %(levelname)-8s %(message)s",
             level=logging.INFO,
@@ -21,17 +22,17 @@ class DistributedTorchRayActor:
         )
         self._world_size = world_size
         self._rank = rank
-        self._local_rank = local_rank
         self._master_addr = master_addr if master_addr else self._get_current_node_ip()
         self._master_port = master_port if master_port else self._get_free_port()
         os.environ["MASTER_ADDR"] = self._master_addr
         os.environ["MASTER_PORT"] = str(self._master_port)
         os.environ["WORLD_SIZE"] = str(self._world_size)
         os.environ["RANK"] = str(self._rank)
-        # NOTE: Ray will automatically set the CUDA_VISIBLE_DEVICES
-        # environment variable for each actor, so always set device to 0
-        # os.environ["LOCAL_RANK"] = str(self._local_rank)
-        os.environ["LOCAL_RANK"] = "0"
+        # NOTE: Ray will automatically set the *_VISIBLE_DEVICES
+        # environment variable for each actor, unless
+        # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES is set, so
+        # set local rank to 0 when the flag is not applicable.
+        os.environ["LOCAL_RANK"] = str(ray.get_gpu_ids()[0]) if ray_noset_visible_devices() else "0"
 
     @staticmethod
     def _get_current_node_ip():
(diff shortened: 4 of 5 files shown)


hijkzzz:
@HollowMan6 https://github.com/OpenRLHF/OpenRLHF/pull/537

HollowMan6:
> @HollowMan6 It is recommended to merge #537 into this MR And only modify the `import modules` related to DeepSpeed. Clean up unused `modules` for other `import xxx`, and place all Ray-related utility in the `ray` directory.

PR #537 actually can be a separate one, as now we no longer call `ray_noset_visible_devices()` in `LLMRayActor`, so something like https://github.com/OpenRLHF/OpenRLHF/pull/524#issuecomment-2497187705 should not happen again even if you only merge this PR.

In addition, It's still unknown what exact import in `openrlhf.utils.deepspeed` has the side effect to make this fail. I also tried `import deepspeed` directly there and it seems like it's not enough to cause this to fail. Also, it makes it challenging and too much overhead to guarantee if the import doesn't undirectly contain things related to deepspeed, e.g., you will never expect this to contain import of DeepspeedStrategy: 

https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/trainer/ray/__init__.py#L1

In fact, having init files empty is usually a sign of better design! Lots of problems are caused by code added to this file, which, in fact, should not be. As a rule of thumb, if any import is done, there is likely a source of problems. You will never expect what your next problem would be.

> Finally, **we should ensure high priority for NVIDIA GPU users**, meaning they should not encounter any errors without modifying the code. That is **NVIDIA GPU users** should not use `export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1`. Thanks!

Yes, there should be no problem, as I mentioned here, https://github.com/OpenRLHF/OpenRLHF/pull/524#issuecomment-2497329923, in any case, it will work just like before, when you don't set the `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES`, as this is a non-intrusive modification. It's unfortunate that no one would expect `from openrlhf.utils.deepspeed import DeepspeedStrategy` can cause this to fail.

Regarding **NVIDIA GPU users** should not use `export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1`, no methods can force users not to do that. If they do, the current upstream codebase is still broken, and this PR is the fix.

hijkzzz:
@HollowMan6 
A question: why execute  `noset_visible_devices = ray_noset_visible_devices(ray.get(get_all_env_variables.remote()))` in `create_vllm_engines` can solve the NVIDIA GPU env issues? 
I’m asking this to understand whether the current approach happens to work by chance, rather than being reliably correct.

HollowMan6:
> @HollowMan6 A question: why only execute `noset_visible_devices = ray_noset_visible_devices(ray.get(get_all_env_variables.remote()))` in `create_vllm_engines` can solve the NVIDIA GPU env issues? I’m asking this to understand whether the current approach happens to work by chance, rather than being reliably correct.

@HollowMan6 It should be reliably correct, since `create_vllm_engines` is not scheduled in any ray actor (also not in `LLMRayActor`) (this can also be confirmed by https://github.com/OpenRLHF/OpenRLHF/pull/524#discussion_r1850330020). Something like https://github.com/OpenRLHF/OpenRLHF/pull/524#issuecomment-2501505023 will only happen when a CUDA call is scheduled in that ray process with `CUDA_VISIBLE_DEVICES`, and that `CUDA_VISIBLE_DEVICES` needs to be changed after that call.

https://github.com/OpenRLHF/OpenRLHF/blob/83d381b77c9d53ed67272464d9acc885ceabe0b8/openrlhf/cli/train_ppo_ray.py#L146-L154

But in any case, something like #537 should be the ultimate guarantee so that we can be sure no `import deepspeed` or any other CUDA calls contaminated the ray actor process when initializing

openrlhf/trainer/ray/vllm_engine.py
```diff
@@ -18,7 +26,8 @@ def __init__(self, *args, **kwargs):
         self.__version__ = vllm.__version__
         assert self.__version__ >= "0.4.1", "OpenRLHF only supports vLLM >= 0.4.1"
 
-        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1
+        noset_visible_devices = kwargs.pop("noset_visible_devices", False)
+        self.use_gpu_executor = kwargs["tensor_parallel_size"] == 1 and not noset_visible_devices
```
hijkzzz, line 30: why we need `not noset_visible_devices` here?
`use_gpu_executor ` is designed for TP=1

openrlhf/trainer/ray/vllm_engine.py
```diff
@@ -83,12 +92,17 @@ def create_vllm_engines(
     max_model_len: int,
 ):
     vllm_engines = []
+    # RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES will always be set in current context,
+    # So we need to get env variables from ray process to check if it is set.
+    noset_visible_devices = ray_noset_visible_devices(ray.get(get_all_env_variables.remote()))
```
hijkzzz, line 97: Why we have to get the env from remote worker?

HollowMan6: I mentioned it here: https://github.com/OpenRLHF/OpenRLHF/pull/524#discussion_r1850330020

`RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` is always set in the ray driver/main process (when executing create_vllm_engines) (by ray), we need to go into a ray actor/function process to figure out if the user actually sets this or not.

HollowMan6: Also in the comment

HollowMan6:
> Another question, so why move `ray_noset_visible_devices` to driver worker can avoid `torch.cuda.is_available()` issue?

Actually, I have already kind of explained it previously:

> It should be reliably correct, since create_vllm_engines is not scheduled in any ray actor (also not in LLMRayActor) (this can also be confirmed by https://github.com/OpenRLHF/OpenRLHF/pull/524#discussion_r1850330020). Something like https://github.com/OpenRLHF/OpenRLHF/pull/524#issuecomment-2501505023 will only happen when a CUDA call is scheduled in that ray actor process with CUDA_VISIBLE_DEVICES pre-set, and that CUDA_VISIBLE_DEVICES needs to be changed after that call.

To make it more clear, `LLMRayActor` will be in a separate process than the one in Ray driver/main process (`create_vllm_engines`) (Of course, a separate context for PyTorch), and vLLM will spawn additional ray actor process to run workers, except for vLLM driver worker, as it will directly use the `LLMRayActor` process, which case the error to happen because that specific process gets contaminated (`torch.cuda.is_available()` is called in that process before ray sets `CUDA_VISIBLE_DEVICES`).

Move `ray_noset_visible_devices` to driver worker can cause no issue, not only because `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` is automatically always set by ray on that process and ray won't modify `CUDA_VISIBLE_DEVICES`, but also because no GPU job will be run directly on the driver worker process.

> for one specific worker that is at the same node of LLMRayActor process, vLLM will not use the actual ray reserved process, but rather update the CUDA_VISIBLE_DEVICES and directly [start the driver worker in LLMRayActor process](https://github.com/vllm-project/vllm/blob/6e9ff050c8e83ad6d5e5eab621e83549e35933a1/vllm/executor/ray_gpu_executor.py#L150-L152). This modification will work if any of the CUDA API hasn't been called, but unfortunately it doesn't because import deepspeed is executed.

hijkzzz:
> > Another question, so why move `ray_noset_visible_devices` to driver worker can avoid `torch.cuda.is_available()` issue?
> 
> Actually, I have already kind of explained it previously:
> 
> > It should be reliably correct, since create_vllm_engines is not scheduled in any ray actor (also not in LLMRayActor) (this can also be confirmed by [#524 (comment)](https://github.com/OpenRLHF/OpenRLHF/pull/524#discussion_r1850330020)). Something like [#524 (comment)](https://github.com/OpenRLHF/OpenRLHF/pull/524#issuecomment-2501505023) will only happen when a CUDA call is scheduled in that ray actor process with CUDA_VISIBLE_DEVICES pre-set, and that CUDA_VISIBLE_DEVICES needs to be changed after that call.
> 
> To make it more clear, `LLMRayActor` will be in a separate process than the one that is Ray driver/main process (`create_vllm_engines`) (Of course, a separate context for PyTorch), and vLLM will spawn additional ray actor process to run workers (for tensor_parallel_size > 1), except for vLLM driver worker (only when `VLLM_USE_RAY_SPMD_WORKER` is not set, as if it's set, there won't exist any driver worker), as it will directly use the `LLMRayActor` process, which case the error to happen because that specific process gets contaminated (`torch.cuda.is_available()` is called in that process before ray sets `CUDA_VISIBLE_DEVICES`).
> 
> Move `ray_noset_visible_devices` to driver worker can cause no issue, not only because `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` is automatically always set by ray on that process and ray won't modify `CUDA_VISIBLE_DEVICES`, but also because no GPU job will be run directly on the driver worker process.
> 
> > for one specific worker that is at the same node of LLMRayActor process, vLLM will not use the actual ray reserved process, but rather update the CUDA_VISIBLE_DEVICES and directly [start the driver worker in LLMRayActor process](https://github.com/vllm-project/vllm/blob/6e9ff050c8e83ad6d5e5eab621e83549e35933a1/vllm/executor/ray_gpu_executor.py#L150-L152). This modification will work if any of the CUDA API hasn't been called, but unfortunately it doesn't because import deepspeed is executed.

could you add me wechat account/ or others through the email janhu9527@gmail.com?

hijkzzz:
New bug
I just set `export RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1` with NVIDIA GPUS
```
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] Error executing method update_weight. This might cause deadlock in distributed execution.
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] Traceback (most recent call last):
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]   File "/home/jianh/.local/lib/python3.10/site-packages/vllm/worker/worker_base.py", line 458, in execute_method
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]     return executor(*args, **kwargs)
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]   File "/tmp/ray/session_2024-11-28_02-19-27_203397_30861/runtime_resources/working_dir_files/_ray_pkg_a9d533e9ca37c04b/openrlhf/trainer/ray/vllm_worker_wrap.py", line 39, in update_weight
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]     torch.distributed.broadcast(weight, 0, group=self._model_update_group)
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]   File "/home/jianh/.local/lib/python3.10/site-packages/torch/distributed/c10d_logger.py", line 83, in wrapper
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]     return func(*args, **kwargs)
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]   File "/home/jianh/.local/lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 2421, in broadcast
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466]     work = group.broadcast([tensor], opts)
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] torch.distributed.DistBackendError: NCCL error in: ../torch/csrc/distributed/c10d/NCCLUtils.hpp:317, unhandled cuda error (run with NCCL_DEBUG=INFO for details), NCCL version 2.21.5
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] ncclUnhandledCudaError: Call to CUDA function failed.
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] Last error:
(LLMRayActor pid=43469) ERROR 11-28 02:28:01 worker_base.py:466] Cuda failure 'invalid argument'
```

PR #537, opened 2024-11-25
Empty all the __init__ files + clean up unused imports

HollowMan6:
Correctness checked by:

```bash
flake8 --select=F401,F821 .
pylint --disable=all --enable=W0611,E0602 .
```

PS: There are a lot of technical debts in the code base, I would suggest maintainers fix them using the linting tools when free.


hijkzzz:
merged 
will ack to @HollowMan6

PR #540, opened 2024-11-26
Use worker_cls when vLLM version > 0.6.4.post1

HollowMan6:
`worker_module_name` and `worker_class_name` is no longer supported.

Refer to https://github.com/vllm-project/vllm/pull/10555


hijkzzz:
vllm will provide a official api to openrlhf

PR #578, opened 2024-12-12
enable overlap_comm in deepspeed

xiaoxigua999:


xiaoxigua999:
The training process does not converge with DeepSpeed v0.15.0 ?
Back line is PPO with overlap_comm = True
image

PR #580, opened 2024-12-12
[WIP] Support training ppo with prm

zhuzilin:
This PR is trying to support PRM within the ppo training.

The main idea is
1. Reconstruct the `Samples` by spliting the response with `prm_step_separator`.
2. Convert the reward sent from remote prm into `list[torch.Tensor]` where the reward is added to the end of the step. 
3. Change the reward (`r`) into 2 types, one for wandb upload (`experience.info["reward"]`), the other for actual training (`experience.reward`) and make the `reward_fn` return both of them.

Known issues:

- Only support one remote PRM at the moment.
- Only support when packing samples is on, but it shouldn't be hard to implement the padding version.

### DIFF
--- a/openrlhf/cli/train_ppo_ray.py
+++ b/openrlhf/cli/train_ppo_ray.py
@@ -1,6 +1,6 @@
 import argparse
 from datetime import datetime
-from typing import List
+from typing import Union
 
 import ray
 import torch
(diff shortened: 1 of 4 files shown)


hijkzzz:
This design is too complex, especially with `generate_stepped_samples` and `reward_fn`. We shouldn't create a solution that forcefully incorporates PRM.

Luc4Gui:
**Post Translation:**

I implemented PRM for PPO training by returning a reward tensor with the same shape as the actor's input IDs. However, when validating my model on Math-Shepherd, I found that the accuracy on the GSM8K test set was only 81.05% (compared to 84.15% reported in the paper). 

Both the SFT model and PRM I used are from [https://huggingface.co/peiyi9979](https://huggingface.co/peiyi9979). For PRM, I calculated `score.weight` by subtracting the weight of the "-" token from the "+" token in the `llm_head`. 

Here are the PPO training parameters I used:
```
--micro_train_batch_size 16 \
--train_batch_size 512 \
--rollout_batch_size 2048 \
--max_epochs 1 \
--prompt_max_len 1024 \
--generate_max_len 1024 \
--zero_stage 3 \
--bf16 \
--actor_learning_rate 1e-7 \
--critic_learning_rate 1e-6 \
--init_kl_coef 0.04 \
--n_samples_per_prompt 1 \
```

The training process logged in WandB is shown in the attached figure (the reward in the log is the sum of step rewards). I would like to ask for advice on what might be causing the issue. I can provide details of the code modifications if necessary.
wandb_log

Luc4Gui:
I implemented PRM for PPO training by returning a reward tensor with the same shape as the actor's input IDs. However, when validating my model on Math-Shepherd, I found that the accuracy on the GSM8K test set was only 81.05% (compared to 84.15% reported in the paper). 

Both the SFT model and PRM I used are from [https://huggingface.co/peiyi9979](https://huggingface.co/peiyi9979). For PRM, I calculated `score.weight` by subtracting the weight of the "-" token from the "+" token in the `llm_head`. 

Here are the PPO training parameters I used:
```
--micro_train_batch_size 16 \
--train_batch_size 512 \
--rollout_batch_size 2048 \
--max_epochs 1 \
--prompt_max_len 1024 \
--generate_max_len 1024 \
--zero_stage 3 \
--bf16 \
--actor_learning_rate 1e-7 \
--critic_learning_rate 1e-6 \
--init_kl_coef 0.04 \
--n_samples_per_prompt 1 \
```

The training process logged in WandB is shown in the attached figure (the reward in the log is the sum of step rewards). I would like to ask for advice on what might be causing the issue. I can provide details of the code modifications if necessary.
wandb_log

openrlhf/cli/train_ppo_ray.py
```diff
@@ -18,8 +18,36 @@
...
+    Returns:
+        Union[torch.Tensor, list[torch.Tensor]]: the reward used for ppo training
+        torch.Tensor: the reward used for logging
+    """
+    orm_rewards = [reward for reward in rewards if not isinstance(reward, list)]
+    prm_rewards = [reward for reward in rewards if isinstance(reward, list)]
+    assert len(prm_rewards) <= 1, "Only support one PRM reward model at the moment."
+
+    if len(prm_rewards) == 1:
+        prm_reward = prm_rewards[0]
```
chaiyixuan, line 32: 1

deema-A:
@Luc4Gui Hi, is there a way to get access to your code? thanks!

PR #586, opened 2024-12-14
Add support for multiturn sft training with packing_samples

UbeCc:
I add support for multiturn sft training with packing_samples. And I will support training without packing_samples soon.

image

The script is,
```bash
set -x

read -r -d '' training_commands <<EOF
openrlhf.cli.train_sft \
   --max_len 2048 \
   --dataset euclaise/gsm8k_multiturn \
   --apply_chat_template \
   --input_key conversations \
   --train_batch_size 16 \
   --micro_train_batch_size 2 \
   --max_samples 20000 \
   --pretrain Qwen/Qwen2.5-1.5B-Instruct \
   --save_path ./checkpoint/test \
   --save_steps -1 \
   --logging_steps 1 \
   --eval_steps -1 \
   --zero_stage 2 \
   --max_epochs 2 \
   --bf16 \
   --flash_attn \
   --learning_rate 5e-6 \
   --load_checkpoint \
   --gradient_checkpointing \
   --packing_samples \
   --multiturn \
   --ring_attn_size 2 \
   --ring_head_stride 2 \
EOF

if [[ ${1} != "slurm" ]]; then
    deepspeed --module $training_commands
fi
```

Currently, the training curve for sft is in line with the usual practice. However, the persuasive evaluation should be testing performance on public benchmarks, which is labor-consuming. I'll gradually complete it.


lsy641:
I have checked the codes. Are they verified? I am worried about the correctness because it calculates the start and end index for each assistant message by tokenizing one message by one message. Then these indexes are applied to the tokenization of all messages. According to [huggingface guideline](https://huggingface.co/docs/trl/en/sft_trainer),   due to lack of context in response_template, the same string (”### Assistant:”) is tokenized differently. So the indexes obtained by tokenizing one by one might not aligned with the results from tokenizing the whole.

UbeCc:
> I have checked the codes. Are they verified? I am worried about the correctness because it calculates the start and end index for each assistant message by tokenizing one message by one message. Then these indexes are applied to the tokenization of all messages. According to [huggingface guideline](https://huggingface.co/docs/trl/en/sft_trainer), due to lack of context in response_template, the same string (”### Assistant:”) is tokenized differently. So the indexes obtained by tokenizing one by one might not aligned with the results from tokenizing the whole.

Currently I've checked using qwen and glm series, but I have not tested other series now. You can try at `sft_trainer.py` by printing the token at `start_idx` and `end_idx`, thx :)

UbeCc:
image

I've tried swe-bench with nebius trajectories [https://huggingface.co/datasets/nebius/SWE-agent-trajectories]. I randomly sample 2k successfully trajectoies and fine-tuned qwen-coder 7b instruct. 

I evaluated the models on SWE-bench Verified after training for 3 epoches. And the experiments are repeated three times.
w/multiturn: {32,29,30}/500 = 6.07%
wo/multiturn: {26,28,28}/500 = 5.47%
Without fine-tuning, qwen-coder 7b instruct performs like 2~3% pass rate.

openrlhf/datasets/sft_dataset.py
```diff
@@ -131,7 +171,7 @@ def __getitem__(self, idx):
             # to avoid EOS_token truncation
             input_token["input_ids"][0][-1] = self.tokenizer.eos_token_id
             input_token["attention_mask"][0][-1] = True
-        info = {"input": prompt, "output": response, "input_length": input_token["attention_mask"].int().sum().item()}
+        info = {"input": prompt, "output": response, "input_length": input_token["attention_mask"].int().sum().item(), "response_ranges": self.response_ranges[idx] if self.multiturn else None}
```
hijkzzz, line 174: It seems we should remove input_length, as it will no longer be used?

UbeCc: Currently, `input_length` is set to be the length of `prompt` + `response` and it is used for `ring-attn` generation (as we should split the current sequence to each computing device). This could be replaced with response_ranges[-1][1] (the final token of the last response theoretically). Should we replace it?

UbeCc: Removing input_length means using response_ranges for all settings. However, getting response_ranges need to shape the dataset to a specific format. For clearity, I think keep input_length is better.

teknium1:
yesyesyesyesyes

PR #608, opened 2024-12-20
Generate prompts from all actors together

zhuzilin:
This PR tries to first send all the prompts to the vllm engine, and then do `llm.generate` all together. In this way, we can maximize the throughput of vllm.


hijkzzz:
A very important optimization, which will be addressed after handling it properly in SGLang + Inference Engine Offload.

hijkzzz:
@ zhaochenyang20

hijkzzz:
this mr has been merged into https://github.com/OpenRLHF/OpenRLHF/pull/677/files
will ack to @zhuzilin

PR #617, opened 2024-12-23
Update LoRA adapter

UbeCc:


hijkzzz:
Merged into main, will ack @UbeCc

PR #648, opened 2025-01-02
Fix: In prm training, we add placeholder token and reward tokens to tokenizer

### LINKED ISSUE #645: Maybe we should add placeholder_token as special token
```
[rank0]:   File "OpenRLHF/openrlhf/datasets/process_reward_dataset.py", line 89, in __getitem__
[rank0]:     labels[input_ids == self.placeholder_token_id] = torch.tensor(label_tokens, dtype=input_ids.dtype)
[rank0]: RuntimeError: shape mismatch: value tensor of shape [5] cannot be broadcast to indexing result of shape [0]
```

LinXueyuanStdio:
fix #645 

image

### DIFF
--- a/examples/scripts/train_prm_llama.sh
+++ b/examples/scripts/train_prm_llama.sh
@@ -2,13 +2,13 @@ set -x
 
 read -r -d '' training_commands <<EOF
 openrlhf.cli.train_prm \
-   --save_path ./checkpoint/mistal-7b-prm \
+   --save_path ./checkpoint/llama3-8b-prm \
    --save_steps 500 \
    --logging_steps 1 \
    --eval_steps 100 \
    --train_batch_size 256 \
    --micro_train_batch_size 8 \
-   --pretrain mistralai/Mistral-7B-v0.1  \
+   --pretrain meta-llama/Llama-3-8B  \
    --bf16 \
    --max_epochs 1 \
    --max_len 8192 \
@@ -23,7 +23,9 @@ openrlhf.cli.train_prm \
    --packing_samples \
    --wandb_group prm \
    --placeholder_token ки \
-   --reward_tokens + -
+   --reward_tokens + - \
+   --specialize_reward_tokens \
+   --placeholder_token_in_tokenizer "<|reserved_special_token_0|>"
 EOF
      # --use_wandb [WANDB_TOKENS] or True (use wandb login command)
      # --packing_samples
--- a/examples/scripts/train_prm_mistral.sh
+++ b/examples/scripts/train_prm_mistral.sh
@@ -23,7 +23,8 @@ openrlhf.cli.train_prm \
    --packing_samples \
    --wandb_group prm \
    --placeholder_token ки \
-   --reward_tokens + -
+   --reward_tokens + - \
+   --specialize_reward_tokens
 EOF
      # --use_wandb [WANDB_TOKENS] or True (use wandb login command)
      # --packing_samples
--- a/openrlhf/cli/train_prm.py
+++ b/openrlhf/cli/train_prm.py
@@ -32,6 +32,22 @@ def train(args):
     )
     # configure tokenizer
     tokenizer = get_tokenizer(args.pretrain, model.model, "right", strategy, use_fast=not args.disable_fast_tokenizer)
+    if args.specialize_reward_tokens:
+        additional_special_tokens = [
+            args.placeholder_token,
+            *args.reward_tokens,
+        ]
+        tokenizer.add_special_tokens({"additional_special_tokens": additional_special_tokens})
+        strategy.print((
+            f"placeholder token {repr(args.placeholder_token)} "
+            f"and reward tokens {repr(args.reward_tokens)} "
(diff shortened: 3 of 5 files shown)


hijkzzz:
The llama/qwen model already have some default special tokens?

LinXueyuanStdio:
Neither.

LinXueyuanStdio:
Correction: Qwen has no reserved special token, while Llama has `<|reserved_special_token_0|>` to `<|reserved_special_token_250|>`

Unfortunately, these reserved tokens are untrained in neither base nor instruct model. The embeddings are set to zero.
Tokens `["ки", "+", "-"]` are trained in Llama. Qwen's `"ки"` remains untrained, while `["+", "-"]` are trained.

I use the following code to detect untrained tokens.
```python
def detect_untrained_tokens(model, eps=1e-16):
    embedding_matrix = model.get_input_embeddings().weight.data
    indicator_untrained = torch.amax(embedding_matrix, dim=1) <= eps
    where_untrained = torch.where(indicator_untrained)[0]
    n_untrained = where_untrained.shape[0]
    untrained_token_ids = where_untrained.detach().numpy().tolist()
    if n_untrained != 0:
        print(f"{n_untrained} untrained tokens detected.")
    return n_untrained, untrained_token_ids
```
Llama results:
image

Qwen results:
image

LinXueyuanStdio:
I decided to keep it simple by ignoring those untrained reserved tokens. When focusing on new special tokens, the best way is to call `model.resize_token_embeddings(new_vocab_size)`

LinXueyuanStdio:
`resize_token_embeddings` will fail when using deepspeed stage 3.

```
[rank1]:   File "OpenRLHF/openrlhf/cli/train_prm.py", line 44, in train
[rank1]:     model.model.resize_token_embeddings(new_vocab_size)
[rank1]:   File "anaconda3/envs/o1/lib/python3.10/site-packages/transformers/modeling_utils.py", line 2117, in resize_token_embeddings
[rank1]:     model_embeds = self._resize_token_embeddings(new_num_tokens, pad_to_multiple_of, mean_resizing)
[rank1]:   File "anaconda3/envs/o1/lib/python3.10/site-packages/transformers/modeling_utils.py", line 2142, in _resize_token_embeddings
[rank1]:     new_embeddings = self._get_resized_embeddings(
[rank1]:   File "anaconda3/envs/o1/lib/python3.10/site-packages/transformers/modeling_utils.py", line 2248, in _get_resized_embeddings
[rank1]:     old_num_tokens, old_embedding_dim = old_embeddings.weight.size()
[rank1]: ValueError: not enough values to unpack (expected 2, got 1)
```

Because transformers can not detect zero3 when HfDeepSpeedConfig is released. I don't understand why to avoid global effects here.

https://github.com/OpenRLHF/OpenRLHF/blob/fb170bc161dc23244eeefe6f9ec04306a61cfd24/openrlhf/models/actor.py#L55-L60


`is_deepspeed_zero3_enabled()` will be false in train_prm.py (even after deepspeed initialized), so we need to handle such a corner case. It becomes a little ugly.

```py
ds_config = strategy.get_ds_train_config(is_actor=True)
if ds_config is not None and ds_config["zero_optimization"]["stage"] == 3:
    from transformers.integrations.deepspeed import HfDeepSpeedConfig
    dschf = HfDeepSpeedConfig(ds_config). # to ensure is_deepspeed_zero3_enabled() == true
model.model.resize_token_embeddings(new_vocab_size)
```

hijkzzz:
@zhuzilin suggests that we should avoid change the model embedding size

LinXueyuanStdio:
Then, to resolve the tokenization issue, another way is to replace the original placeholder text with a reserved trained token. It is easy to implement. However, such a replacement setting should also be transferred to evaluation and deployment. People may forget and make mistakes.

Do you have any good ideas?

hijkzzz:
> Then, to resolve the tokenization issue, another way is to replace the original placeholder text with a reserved trained token. It is easy to implement. However, such a replacement setting should also be transferred to evaluation and deployment. People may forget and make mistakes.
> 
> Do you have any good ideas?

Users should customize these tokens based on their own models instead of using the default ones.

LinXueyuanStdio:
If the dataset has only 'ки' or only '_ки', we can actually deduce the real token id of the placeholder token. This rule fails when there are both.

The most secure way is to process the text.

LinXueyuanStdio:
Ok, maybe I will add a '--placeholder_token_reserved_in_tokenizer' option.

LinXueyuanStdio:
I add '--placeholder_token_in_tokenizer' and '--reward_tokens_in_tokenizer' options.

Original placeholder_token in the dataset will be replaced with placeholder_token_in_tokenizer. If placeholder_token_in_tokenizer is not provided by user, it will be initialized with placeholder_token by default.

See the sample below, where placeholder_token=='ки' and placeholder_token_in_tokenizer == '<|reserved_special_token_0|>'.

image

LinXueyuanStdio:
The question is more about the data format. I would like to close this PR.

PR #649, opened 2025-01-02
fix: initialize loss_mean with first batch loss

Freder-chen:
This PR aligns the `loss_mean` calculation in the trainer to ensure consistent behavior. Notably, the PPO trainer does not seem to exhibit this issue.

### Expected Behavior

- The `loss_mean` is initialized using loss value from the first training step.
- For subsequent steps, the `loss_mean` is updated using EMA.

### Known Issue

- **Resuming Training**: When training is interrupted and resumed, the `loss_mean` may not retain its accurate value since it is not saved in checkpoint. This could lead to inconsistencies in EMA calculation during resumed training.


hijkzzz:
fixed in https://github.com/OpenRLHF/OpenRLHF/pull/650

PR #650, opened 2025-01-03
Fix loss_mean log

xiaoxigua999:


openrlhf/trainer/dpo_trainer.py
```diff
@@ -177,16 +177,14 @@ def fit(self, args, consumed_samples=0, num_update_steps_per_epoch=None):
                 self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler)
 
                 acc = (chosen_reward > reject_reward).float().mean().item()
-                acc_mean = acc_mean * 0.9 + 0.1 * acc
-                loss_mean = loss_mean * 0.9 + 0.1 * preference_loss.item()
+                acc_sum += acc
+                loss_sum += preference_loss.item()
```
Freder-chen, line 181: The `loss_sum` should be placed in the outer loop; otherwise, anomalous behavior may occur at the end of an epoch.

hijkzzz: `loss_sum = 0` is OK?

PR #661, opened 2025-01-09
[WIP] Integrate SGLang into OpenRLHF

zhaochenyang20:
This PR supports SGLang as an inference engine for the PPO actor and makes relevant changes. Detailed usage can be found in:

https://github.com/zhaochenyang20/Awesome-ML-SYS-Tutorial/blob/main/rlhf/OpenRLHF/openrlhf-sglang.md


openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -662,41 +661,67 @@ def _generate_vllm(self, all_prompts: List[str], **kwargs) -> List[Samples]:
...
+                    llm.generate.remote(
+                        sampling_params=sampling_params, prompt_token_ids=prompt_token_ids, all_prompts=all_prompts
+                    )
                 )
 
         # Retrieve and combine results from all outputs
         all_outputs = sum(ray.get(all_output_refs), [])
+        assert len(all_outputs) == len(all_prompts)
+        pad_token_id, eos_token_id = self.tokenizer.pad_token_id, self.tokenizer.eos_token_id
+        try:
```
fzyzcjy, line 678: nit: it may be a bit better if we do not use try-except because

* if sglang/vllm's format change in the future, non-error may become error and vise versa, and the behavior suddenly change
* `except` without specifying exception type will cause everything to be caught here, even things like real bugs etc

fzyzcjy (commented):
Since @zhaochenyang20 asked me at slack to have a review about this, I very briefly glanced at the code (indeed randomly picked a file, experience_maker.py) but too tired to continue today.

Anyway there is a small nit below, and also a small proposal https://github.com/sgl-project/sglang/issues/2818 that may be hopefully a little bit helpful.

openrlhf/trainer/ppo_utils/experience_maker.py
zhaochenyang20: I also do not like try except, but do I have other choices? Like seeing the type of the first output?

fzyzcjy: I guess a way may be check `if config.engine == 'sglang'`, or maybe `if inference_engine.get_mode() == 'sglang'`.

catqaq:
great job!

openrlhf/cli/batch_inference.py
```diff
@@ -13,20 +13,112 @@
...
-    # configure strategy
+    # Configure strategy
     class Empty:
         pass
 
-    dummy_strategy = Empty()
-    dummy_strategy.print = print
-    dummy_strategy.is_rank_0 = lambda: True
-    dummy_strategy.args = args
+    strategy = Empty()
```
hijkzzz, line 25: Just set strategy to None for vllm and sglang
btw, for reward model please use the deepspeed strategy

openrlhf/trainer/ppo_utils/experience_maker.py
hijkzzz: Yes, this part should be refactored.

openrlhf/cli/batch_inference.py
zhaochenyang20: I just copied from the main.

zhaochenyang20:
delete this print `print("os.environ['LOCAL_RANK']", os.environ["LOCAL_RANK"])`

zhaochenyang20:
split sglang and vllm into two files and provide different `create_inference_engines` funciton.

zhaochenyang20:
delete this `torch.cuda.synchronize()`

merrymercy:
@xiaoxigua999 @hijkzzz This PR has been fully verified regarding accuracy and speed. Is it possible to merge this?

PR #711, opened 2025-02-05
Add support for max time per run

titu1994:
# Changelog
- Adds support for max time per run for instances where compute cluster has a maximum duration for a job.
- Calculates checkpoint at current step, then exits the script. (An option can be to raise KeyboardException, or simply pass signal.SIGINT, if deepspeed supports it - but i haven't verified this).

### DIFF
--- a/openrlhf/utils/utils.py
+++ b/openrlhf/utils/utils.py
@@ -91,7 +91,7 @@ def blending_datasets(
                 eval_data = data[eval_split].select(range(min(max_count, len(data[eval_split]))))
             # train will contains eval? TODO
             else:
-                eval_data = train_data.select(range(min(max_count, int(len(train_data) * 0.03))))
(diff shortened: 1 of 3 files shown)


openrlhf/utils/utils.py
```diff
@@ -91,7 +91,7 @@ def blending_datasets(
                 eval_data = data[eval_split].select(range(min(max_count, len(data[eval_split]))))
             # train will contains eval? TODO
             else:
-                eval_data = train_data.select(range(min(max_count, int(len(train_data) * 0.03))))
+                eval_data = train_data.select(range(min(max_count, max(1, int(len(train_data) * 0.03)))))
```
titu1994, line 94: This is an edge case when the number of samples in the train set is small enough (say for debugging), that the 0.03 * train_data ~ 0 so there are no samples in the eval set and a subsequent step crashes.

PR #721, opened 2025-02-08
Fix vLLM instance scheduling when tp size is 1

HollowMan6:
https://github.com/OpenRLHF/OpenRLHF/pull/677 introduced a change that forces to use ray as the vLLM distributed_executor_backend even if tp size is 1. While this works for vLLM versions below 0.7.0, it's no longer the case starting from 0.7.0, as it will force you to use uni even if you set explicitly to use ray (it takes me so much time to debug this and finds out this weird change XD, as there is no warning information regarding this at all!):

https://github.com/vllm-project/vllm/blob/ad34c0df0f1b26b303a590133685b29e3daad20e/vllm/engine/llm_engine.py#L444-L462

So, we still need to treat this the same way as what we do before for GPU executor, otherwise, if we have multiple vLLM instances scheduled on one machine, it will all use GPU 0, which will cause errors.

In addition, now we specify explicitly to use uni when tp size is 1, in case vLLM changed their mind and moved from this weird behavior later.

### DIFF
--- a/openrlhf/trainer/ray/vllm_engine.py
+++ b/openrlhf/trainer/ray/vllm_engine.py
@@ -22,10 +22,11 @@ def get_all_env_variables():
 class LLMRayActor:
 
     def __init__(self, *args, bundle_indices: list = None, **kwargs):
-        # a hack to make the script work.
-        # stop ray from manipulating CUDA_VISIBLE_DEVICES
-        # at the top-level
-        del os.environ["CUDA_VISIBLE_DEVICES"]
+        if kwargs.get("distributed_executor_backend") == "ray":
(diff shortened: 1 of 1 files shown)


youkaichao:
sorry for the trouble! please let me know if https://github.com/vllm-project/vllm/pull/12934 works for you 🙏

HollowMan6:
Thank you for the quick fix @youkaichao! This PR combined with #725 should fix the issue on the OpenRLHF side already even without https://github.com/vllm-project/vllm/pull/12934, but it's always a good idea to respect `distributed_executor_backend` when `world_size`=1, as not respecting this made things tough to debug. Anyway, thanks again!

PR #723, opened 2025-02-08
Change the vLLM sleep level default to 2

HollowMan6:
According to https://github.com/vllm-project/vllm/pull/11743/files#diff-1661f61c5087d7770dd252ba3fbe544871ec5554a38f247d189ba25343770232R1141-R1151, level 1 will offload the model weights to CPU, but this is not needed in our use case and can be simply discarded here, since we will load the new weight in `update_weight` later when we `_broadcast_to_vllm`.


hijkzzz:
level 2 will discrad rope pytorch buffer so .... we only use level 1 now

HollowMan6:
Ah, good to know!

PR #725, opened 2025-02-08
Fix hybrid engine when vLLM tp size is 1

HollowMan6:
Sorry that I forgot to fix for the hybrid engine scheduling when vLLM tp size is 1 in #721

### DIFF
--- a/openrlhf/trainer/ray/vllm_engine.py
+++ b/openrlhf/trainer/ray/vllm_engine.py
@@ -27,12 +27,12 @@ class LLMRayActor:
             # stop ray from manipulating CUDA_VISIBLE_DEVICES
             # at the top-level when the distributed_executor_backend is ray.
             del os.environ["CUDA_VISIBLE_DEVICES"]
-        # every worker will use 0.2 GPU, so that we can schedule
-        # 2 instances on the same GPUs.
-        if bundle_indices is not None:
-            os.environ["VLLM_RAY_PER_WORKER_GPUS"] = "0.2"
-            os.environ["VLLM_RAY_BUNDLE_INDICES"] = ",".join(map(str, bundle_indices))
-            print(f"creating LLM with bundle_indices={bundle_indices}")
+            # every worker will use 0.2 GPU, so that we can schedule
+            # 2 instances on the same GPUs.
+            if bundle_indices is not None:
+                os.environ["VLLM_RAY_PER_WORKER_GPUS"] = "0.2"
+                os.environ["VLLM_RAY_BUNDLE_INDICES"] = ",".join(map(str, bundle_indices))
+                print(f"creating LLM with bundle_indices={bundle_indices}")
 
         # Number of actors that will send prompt to this engine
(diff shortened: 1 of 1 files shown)


hijkzzz:
This bug has been fixed in: https://github.com/vllm-project/vllm/pull/12934 ?

youkaichao:
you can try `pip install vllm --pre --extra-index-url https://wheels.vllm.ai/nightly` to install the latest wheel from vllm main to see if it's fixed .

HollowMan6:
> This bug has been fixed in: [vllm-project/vllm#12934](https://github.com/vllm-project/vllm/pull/12934) ?

For the current OpenRLHF codebase itself, we still need this PR as we now set to use `uni` when tp=1 for #721: https://github.com/OpenRLHF/OpenRLHF/pull/721/files#diff-d7d5aa9db07c0a235af959a88c3fe0c5440e7441250f9b00009735ba8d0837e6R121. In any case, I think it's better to keep things in this way to keep support for 0.7.0 - 0.7.2.

> you can try pip install vllm --pre --extra-index-url https://wheels.vllm.ai/nightly to install the latest wheel from vllm main to see if it's fixed.

This can be tested when #721 is not merged, but now https://github.com/vllm-project/vllm/pull/12934 is not necessary to get this working.

hijkzzz:
reverted.
because this mr causes tp=2 failed.

HollowMan6:
> reverted. because this mr causes tp=2 failed.

Oh really, I didn't notice any failure when I submitted this PR, and now I tested again with this PR, and there are still no errors, either tp size 1, tp size 2, or even tp size 4. I was using a single node with 4 GPUs for testing,

```
   --ref_num_nodes 1 \
   --ref_num_gpus_per_node 4 \
   --reward_num_nodes 1 \
   --reward_num_gpus_per_node 4 \
   --critic_num_nodes 1 \
   --critic_num_gpus_per_node 4 \
   --actor_num_nodes 1 \
   --actor_num_gpus_per_node 4 \
   --vllm_num_engines 2 \
   --colocate_all_models
``

It's unlikely that this PR is the root cause for tp=2 to fail, as there is no added logic here when the tp size is greater/equal to 2. Setting `placement_group_bundle_index` to `-1` is just to follow the default value.

I'm not sure if you accidentally set `actor_num_nodes * actor_num_gpus_per_node` to be smaller than `vllm_num_engines * vllm_tensor_parallel_size` while testing this PR. If so, then the error is caused by the nature of the current design, and you should end up with errors like these: https://github.com/OpenRLHF/OpenRLHF/pull/728. I helped add an assertion there to ask users to correct those parameters.

If it's something else, please let me know. However, in any case, the current codebase is broken for hybrid engine vLLM tp=1.

HollowMan6:
> reverted. because this mr causes tp=2 failed.

OK, I was able to find the bug for tp=2, when I tested the hybrid engine with 6 nodes. Here is the fix: #732. But this PR is not the root cause, since `placement_group_bundle_index` wasn't set from the beginning (#677) and remains in the current main.

PR #729, opened 2025-02-09
support cuda ipc in hybrid engine

xiaoxigua999:


HollowMan6 (approved):
Tested on AMD machines, and still worked like a charm!

PR #759, opened 2025-02-19
Pack vLLM engines together when tp>1 in distributed RLHF

HollowMan6:
Rewrite the `create_vllm_engines` logic regarding the placement group, so that we create a big placement group for all the vLLM engines, and specify the strategy as `PACK`, to avoid the engines being spread all around the nodes and minimize the communication overhead when tp>1.

Note that now we will force to use vLLM 0.7.2 in any case (either hybrid engine or distributed): https://github.com/vllm-project/vllm/commit/bc1bdecebf76cca0dfafe4924d529b30c8a24795

I believe there should be performance improvements that come with this change:
- Less cross-node communication in vLLM weight update.
- Critic/Reward deepspeed instances get "real" packed, thus improving the critic training speed in PPO.

Before this change:

Test resource allocation parameters on 6 nodes:
```
   --ref_num_nodes 2 \
   --ref_num_gpus_per_node 4 \
   --reward_num_nodes 2 \
   --reward_num_gpus_per_node 4 \
   --critic_num_nodes 2 \
   --critic_num_gpus_per_node 4 \
   --actor_num_nodes 2 \
   --actor_num_gpus_per_node 4 \
   --vllm_num_engines 4 \
   --colocate_critic_reward \
   --colocate_actor_ref
```

vLLM workers get spread all over the nodes (4/6 nodes look like this), which also stops the Critic/Reward from getting packed together:
Screen20250219-143712

### DIFF
--- a/openrlhf/trainer/ray/vllm_engine.py
+++ b/openrlhf/trainer/ray/vllm_engine.py
@@ -1,5 +1,4 @@
 import os
-import numpy as np
 from typing import Any, List
 
 import ray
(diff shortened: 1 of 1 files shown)


jovany-wang (approved):

HollowMan6:
Rebased to fix the merge conflict

PR #781, opened 2025-02-24
Fix SFT packing_samples

wangcho2k:
This PR fixes the two issues on SFT training w/ `packing_samples`

### 1. incorrect response_ranges accumulation

Since each `response_ranges` is based on the beginning of the sample (0), the sum of sample lengths up to the previous sample (instead of accum. of previous `response_ranges`) should be added.

Before the fix, the `response_ranges` were completely unaligned to actual responses (even contains previous instruction).

I checked the patch with following snippet:

```python
import os
from dataclasses import dataclass

import torch
from datasets import load_from_disk
from transformers import AutoTokenizer

from openrlhf.datasets import SFTDataset
from openrlhf.utils import get_strategy


@dataclass
class Args:
    input_key: str = "conversations"  # as in test dataset
    zero_stage: str = "2'"
    apply_chat_template: bool = True


args = Args()

os.environ["WORLD_SIZE"] = "1"
os.environ["RANK"] = "0"
os.environ["MASTER_ADDR"] = "127.0.0.1"
os.environ["MASTER_PORT"] = "29500"

strategy = get_strategy(args)
torch.distributed.init_process_group()

tok = AutoTokenizer.from_pretrained("/path/to/tokenizer")
ds = load_from_disk("/path/to/SOME_TEST_SFT_DATASET")

train_dataset = SFTDataset(ds["train"], tok, 512, strategy, multiturn=True)

inputs = [train_dataset[i] for i in range(256, 277)]

# torch.set_printoptions(threshold=10_000)
print("Very first input:")
print(inputs[0])

# verifying packed inputs
packed = train_dataset.packing_collate_fn(inputs)
print("Packed input tokens:")
print(packed[1])
print(packed[1].size())
print("Response Ranges:")
print(packed[3])
print("Response Ranges as text:")
for prompt_range in packed[3]["response_ranges"]:
    print(tok.decode(packed[1][0][prompt_range[0][0] : prompt_range[0][1] + 1]))

```

### 2. drops the last elements in label mask calculation/extraction

Tensor slicing in the label masking/extraction logic ignored the last elements, which are mainly EOS tokens or the last token before the response generation.
Before fix, the trained model cannot generate EOS token, resulting in endless generation (up to `max_token_embeddings`)


hijkzzz:
@UbeCc

hyunwoongko:
I've tested original code but performance was degraded when I used sequence packing.
And this PR resolved the problem.

UbeCc:
> I've tested original code (current main branch) but performance was decreased when I used sequence packing. And this PR resolved the problem. After applying this PR, performance is almost same with non packing results. FYI, I evaluated models on benchmarks like MT-bench, Wildbench, ArenaHard, and etc.

Thank you for you contribution! Can you please provide some more detailed results if possible?

hyunwoongko:
## Results
I used our internal model and dataset:
https://www.arxiv.org/abs/2502.18934

SFT with non-packing + non-ring-attn: 
- MT-bench: 5.9 
- Wildbench: 13.42
- Arena Hard: 13.53

SFT with packing + non-ring-attn: 
- MT-bench: 4.8
- Wildbench: -3.98
- Arena Hard: 6.94

SFT with packing + ring-attn: 
- MT-bench: 4.8563
- Wildbench: -4.47
- Arena Hard: 8.28

Apply this PR with packing + ring-attn:
- MT-bench: 5.9
- Wildbench: 13.57
- Arena Hard: 15.70

---

## Why this problem happend

스크린샷 2025-03-10 오후 3 09 28

I set micro batch size per device as 1 to explain easily, the above is non-packing result and the bottom one is packing result.

You can see the last element of label is -100. This is because this line makes the next token of response masked.
https://github.com/OpenRLHF/OpenRLHF/blob/3de23340eff20982f3fdbf2d70b8d87a0785c51c/openrlhf/datasets/sft_dataset.py#L136

But generally we use tokens like `<|eot_id|>` `<|im_end|>` `<|im_start|>` to distinguish turns. And training 'end of turn' to model is really important. 

And after applying this PR, you can see the result is exactly same with non-paking version.

스크린샷 2025-03-10 오후 3 27 30

hyunwoongko:
OP also mentioned this problem like:

Tensor slicing in the label masking/extraction logic ignored the last elements, which are mainly EOS tokens or the last token before the response generation.
Before fix, the trained model cannot generate EOS token, resulting in endless generation (up to max_token_embeddings)

UbeCc (approved):

UbeCc:
> OP also mentioned this problem like:
> 
> Tensor slicing in the label masking/extraction logic ignored the last elements, which are mainly EOS tokens or the last token before the response generation. Before fix, the trained model cannot generate EOS token, resulting in endless generation (up to max_token_embeddings)

ic and that is the problem. Thank you for your detailed training results!

PR #797, opened 2025-02-26
Add k2_loss

LYMDLUT:
if args.use_kl_loss:
        if args.kl_estimate not in ["k2", "k3"]:
            warnings.warn(f"Recommend setting {args.kl_estimate} to 'k2' or 'k3' when using KL as a loss")
    else:
        if args.kl_estimate not in ["k1"]:
            warnings.warn(f"Recommend setting {args.kl_estimate} to 'k1' when not using KL as a loss.")

### DIFF
--- a/openrlhf/cli/train_ppo_ray.py
+++ b/openrlhf/cli/train_ppo_ray.py
@@ -70,12 +70,7 @@ def train(args):
     vllm_engines = None
     if args.vllm_num_engines is not None and args.vllm_num_engines > 0:
         max_len = args.max_len if args.max_len else args.prompt_max_len + args.generate_max_len
-        if args.colocate_all_models and args.vllm_gpu_memory_utilization >= 0.9:
-            args.vllm_gpu_memory_utilization = 0.4
-            print(
-                f"Set args.vllm_gpu_memory_utilization to {args.vllm_gpu_memory_utilization} for colocate_all_models!"
-            )
-
+        if args.colocate_all_models:
             assert (
                 args.actor_num_nodes * args.actor_num_gpus_per_node
                 == args.vllm_num_engines * args.vllm_tensor_parallel_size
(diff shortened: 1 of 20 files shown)


openrlhf/cli/train_ppo_ray.py
```diff
@@ -42,6 +42,11 @@ def _validate_args(args):
             actor_world_size % critic_world_size == 0
         ), f"actor_world_size must be divisible by critic_world_size, got {actor_world_size} and {critic_world_size}"
 
+    if args.use_kl_loss:
```
hijkzzz, line 45: remove these lines

openrlhf/models/utils.py
```diff
@@ -19,15 +19,31 @@ def compute_approx_kl(
         log_probs_base: Log probabilities of the base distribution.
         action_mask: Mask for actions.
     """
-
-    log_ratio = log_probs.float() - log_probs_base.float()
-    if action_mask is not None:
-        log_ratio = log_ratio * action_mask
-
+    # The relationship and distinction between the KL estimator and
+    # KL loss can be found at: https://zhuanlan.zhihu.com/p/26077220144
```
hijkzzz, line 23: do not use zhihu link (only chinese)

openrlhf/cli/train_ppo_ray.py
richardodliu, line 45: if args.use_kl_loss:
        assert args.kl_estimate in ["k2", "k3"]

openrlhf/cli/train_ppo_ray.py
```diff
@@ -42,6 +42,10 @@ def _validate_args(args):
...
+        assert args.kl_estimate in ["k2", "k3"], f"When using KL as loss, `kl_estimate` must be 'k2' or 'k3'."
+    else:
+        assert args.kl_estimate in ["k1"], f"When not using KL as loss, `kl_estimate` must be 'k1'."
```
richardodliu, line 48: As we only have three choices, I think there is no reason for checking when "else"

hijkzzz: We should not restrict users' choices, such as using KL reward + K2.

PR #809, opened 2025-03-01
fix ring attn in create_vllm_engines

ChaosCodes:
When ring_attn is enable, only `args.actor_num_nodes * args.actor_num_gpus_per_node // args.ring_attn_size` actor will call `generate_samples` in make_experience_list.

The existed code will have a large `self.num_actors` which make the vllm can not get into `if self.actor_counter == self.num_actors:` in `add_requests`

### DIFF
--- a/openrlhf/cli/train_ppo_ray.py
+++ b/openrlhf/cli/train_ppo_ray.py
@@ -88,7 +88,7 @@ def train(args):
             args.enable_prefix_caching,
             args.enforce_eager,
             max_len,
-            args.actor_num_nodes * args.actor_num_gpus_per_node,
+            args.actor_num_nodes * args.actor_num_gpus_per_node // args.ring_attn_size,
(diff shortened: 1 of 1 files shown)


ChaosCodes:
However, when self.num_actors > 1, using ring_attn will have some synchronization issue, I suspect that it is because of the code below
```
# Make sure all requests are sent.
        if self.strategy.ring_attn_group is None:
            torch.distributed.barrier()
 ```

I have no any good idea to solve this

hijkzzz:
could you help to deubg this hang bug?

ChaosCodes:
> could you help to debug this hang issue?

@hijkzzz Sure, I can help with this. But I have tried to only torch.distributed.barrier() on process with `ring rank = 0`, which is tricky and might not work. Do you have any idea that I can have a try?

PR #810, opened 2025-03-01
Fix kl_estimator params config

HollowMan6:
Related to #797

The codebase is currently broken because of the errors introduced.


hijkzzz:
Fixed

PR #836, opened 2025-03-06
added LoRA adapter disabling for computing KL divergence in single no…

wilkincr:
…de PPO. this removes need to store SFT model in exchange for slight overhead of adapter toggling


wilkincr:
I had to add --vllm_enable_sleep to train_ppo.py in order to run, probably not expected behavior but didn't want to modify too much

PR #853, opened 2025-03-11
Fix temperature alignment

dingyuan-shi:
There exsits a misalignment between the log_probs of rollout sampling and log_probs for kl-div.
When conduct PPO training, the temperature can be specified for rollout sampling.
However, the log_probs used for KL-div calculation are directy calculated based on logits and ignores the temperature, which incurs misalignment.
This PR fix this misalignment.


hijkzzz:
Excellent fix, please modify `train_ppo_ray.py` similarly and resolve conflicts. 
Thank you.

PR #869, opened 2025-03-14
support lora+vllm

TideDra:
This PR enables lora with vllm.
When broadcasting to vllm, it firstly merges the lora adapter of a module before broadcasting the parameters of the modue.

### DIFF
--- a/openrlhf/cli/train_dpo.py
+++ b/openrlhf/cli/train_dpo.py
@@ -27,6 +27,7 @@ def train(args):
         lora_alpha=args.lora_alpha,
         lora_dropout=args.lora_dropout,
         target_modules=args.target_modules,
+        exclude_modules=args.exclude_modules,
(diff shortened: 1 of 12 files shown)


TideDra:
@hijkzzz hello, can you have a look at this PR? I think this is an important feature and is widely required by the community. It is already used in lmm-r1

chuangzhidan:
> @hijkzzz hello, can you have a look at this PR? I think this is an important feature and is widely required by the community. It is already used in lmm-r1

一个底膜，多个adapter，为了节省显卡内存？

cjrrr:
when using lora
   --lora_rank 64 \
   --lora_alpha 16 \
   --lora_dropout 0.05 \
   --target_modules q_proj  v_proj 
there is still information about updating weight of the basemodel like this,is that ok?
微信图片_20250723022946

PR #882, opened 2025-03-18
remove ppo trainer.py

hijkzzz:


haoxufd:
why to remove this cli?

PR #907, opened 2025-03-24
Fix: inconsistent vllm performance when tp > 1

whksmo:
During training with vllm tp > 1, inconsistent performance among vLLM engines was observed. Some engines ran significantly slower than others (x2~x4 slower). Looks like the barrier operations in ppo_actor affects the tp communication of vllm engines.
This can be fixed by moving barrier from ppo_actor to vllm_engine.

### DIFF
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -764,12 +764,6 @@ class RemoteExperienceMaker(BaseExperienceMaker):
             )
         ray.get(refs)
 
-        # Make sure all requests are sent.
-        if self.strategy.ring_attn_group is None:
-            torch.distributed.barrier()
-        else:
-            time.sleep(3)
-
         # Retrieve and combine results from all outputs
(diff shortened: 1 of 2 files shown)


tangzuoli:
Please re-examine this PR, as I believe it may lead to a deadlock.  
Consider the following scenario: 4 Actor models colocated with 1 VLLM Engine, where all models are colocated.  
The Actor models on GPUs 0, 1, 2, and 3 respectively add requests. Ray schedules the `add requests` from GPU 0 to execute first, and it returns immediately. At this point, the Actor on GPU 0 will immediately execute `get_responses`. Due to the uncertainty of Ray's scheduling, it is possible that the `get_responses` call from the Actor model on GPU 0 could be processed by the `vllm_actor` before the `add requests` calls from the other GPUs.  

In this PR, this situation would result in an infinite `while` loop because the `vllm_actor` itself can only handle one invocation at a time. The `add requests` from the other GPUs will remain waiting until the `get_responses` call has been fully executed.

PR #947, opened 2025-04-07
Offload lp_grads and lp_params as well for deepspeed >= 0.16.5

HollowMan6:
Those offload types are fixed in https://github.com/deepspeedai/DeepSpeed/pull/7050, and they have already been released in 0.16.5.

### DIFF
--- a/openrlhf/utils/deepspeed/deepspeed_utils.py
+++ b/openrlhf/utils/deepspeed/deepspeed_utils.py
@@ -123,17 +123,25 @@ def offload_deepspeed_states(model, pin_memory=True, non_blocking=True):
         raise NotImplementedError("Only Zero stage 3 is currently supported")
 
     # if zero_stage == 3 and not adam_offload:
+    import deepspeed
     import torch
     from deepspeed.runtime.zero.offload_config import OffloadDeviceEnum, OffloadStateTypeEnum
 
+    offload_state_types = [
+        OffloadStateTypeEnum.optim_states,
+        OffloadStateTypeEnum.contiguous_grad_buffer,
+        OffloadStateTypeEnum.hp_params,
+    ]
+
+    if deepspeed.__version__ >= "0.16.5":
+        # These offload types are fixed in https://github.com/deepspeedai/DeepSpeed/pull/7050
+        offload_state_types += [
+            OffloadStateTypeEnum.lp_grads,
+            OffloadStateTypeEnum.lp_params,
+        ]
+
     model.optimizer.offload_states(
-        include=[
-            OffloadStateTypeEnum.optim_states,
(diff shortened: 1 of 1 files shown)


hijkzzz:
One issue is that when calling `update_weights` of vllm, the actor's parameters must be on the GPU.

HollowMan6:
> One issue is that when calling `update_weights` of vllm, the actor's parameters must be on the GPU.

That's interesting. I didn't experience any errors with this PR, maybe those offloads are not successful even when specified? I will double-check.

hijkzzz:
Thank you for your discovery. I just thought of another issue — when enabling reward ref offload *together with* DeepSpeed sleep, it seems like we should also offload the parameters in this way. 
Also, we could support allowing users to choose whether to offload just the optimizer or include the parameters as well. For example, when `deepspeed_sleep_level = 1`, we offload only the optimizer; when it's set to `2`, we offload everything.

HollowMan6:
> When enabling reward ref offload _together with_ DeepSpeed sleep, it seems like we should also offload the parameters of ref /reward in this way.

Just did another investigation and it looks like DeepSpeed hasn't added support for the dynamic offload in pure inference mode (when the optimizer is not specified).

When the optimizer is not specified, the optimizer will be type `DeepSpeedZeRoOffload` instead of `DeepSpeedZeroOptimizer_Stage3`. Reference: https://github.com/deepspeedai/DeepSpeed/blob/56005d2b256eb81a88cba0a1984375f9663a3110/deepspeed/runtime/engine.py#L1684-L1707

```log
ray.exceptions.RayTaskError(AttributeError): ray::ReferenceModelRayActor.init_model_from_pretrained()
  File "openrlhf/trainer/ray/launcher.py", line 134, in init_model_from_pretrained
    self.offload_states()
  File "openrlhf/trainer/ray/launcher.py", line 106, in offload_states
    offload_deepspeed_states(model)
  File "openrlhf/utils/deepspeed/deepspeed_utils.py", line 143, in offload_deepspeed_states
    model.offload_states(
  File "deepspeed/runtime/engine.py", line 3904, in offload_states
    self.optimizer.offload_states(include=include, device=device, pin_memory=pin_memory, non_blocking=non_blocking)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DeepSpeedZeRoOffload' object has no attribute 'offload_states'
```

> Also, we could support allowing users to choose whether to offload just the optimizer or include the parameters as well. For example, when `deepspeed_sleep_level = 1`, we offload only the optimizer; when it's set to `2`, we offload everything.

It looks like parameters offload is disabled https://github.com/OpenRLHF/OpenRLHF/commit/ac2f0a28a7db6df7be1013265e2b534baabec860, so I guess this is not needed now. If not, this should be straightforward to implement.

PR #968, opened 2025-04-15
support dynamic sampling

TideDra:
This PR implement dynamic sampling in [DAPO](http://arxiv.org/abs/2503.14476).
Important features:
- `num_episode` is replaced with `max_global_steps`. Since dynamic sampling may skip the training stage in a loop in order to accumulate enough experiences, the real training steps is less than the expected steps calculated by `num_episode`.
- Use `custom_experience_filter` to filter expericence, just as `custom_reward_fn`
- Oversampled experience can be stored in replay_buffer and used in next step


hijkzzz:
dynamic sampling should be implemented by all_gather_object_list -> apply filter -> scatter_object_list ?

TideDra:
> dynamic sampling should be implemented by all_gather_object_list -> apply filter -> scatter_object_list ?

gather/scatter can be done in the customized filter, if the user need it

hijkzzz:
the single controller will be merged https://github.com/OpenRLHF/OpenRLHF/pull/972
so that we can implemented adpo based on  single controller

PR #989, opened 2025-04-23
Merge lmm-r1 for Multimodal PPO

TideDra:
Breaking changes:
- Support Qwen2.5-VL, Phi3.5/4
- Easily customize new model
- Integrate dynamic sampling #968 
- Integrate lora+vllm #869 
- prompt with visual information is required to be a stringfied message. Since DataProcessor always apply chat-template, `--apply_chat_template` is deprecated, and stringfied message should be the only unified data format in the future.

### DIFF
--- a/examples/scripts/experience_filter.py
+++ b/examples/scripts/experience_filter.py
@@ -1,0 +1,4 @@
+import torch
+
+def experience_filter(experience_maker,experiences):
+    return experiences[: len(experiences) // 2]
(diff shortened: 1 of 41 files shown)


hijkzzz:
This is a very good MR. Currently, OpenRLHF is focused on the development of DeepSpeed AutoTP. This MR will be refactored later. The plan is as follows:

1. Refactor the multimodal code, mainly to make it easier to understand without compromising the readability of the text module.
2. Other features, such as LoRA and VLLM, can be temporarily left out. The reason is that we need to simplify and streamline the framework step by step.

TideDra:
Thanks for your review. Let me know if there is anything I can do.

PR #1002, opened 2025-04-28
Support specifying ds tp size separately for each model

HollowMan6:
Especially for PPO, it can be very useful to have separate tp size to make it more flexible.

Now we still have `--ds_tensor_parallel_size` if users don't bother setting it for each model (also for compatibility), and we have `--xxx_tensor_parallel_size` to make the configuration overridable at specific model level.


hijkzzz:
翻译如下：

"This feature is very meaningful for async RL, but it seems that `setup_distributed` does not pass in the correct `ds_tp`, and the logic is a bit convoluted?"

HollowMan6:
> This feature is very meaningful for async RL, but it seems that `setup_distributed` does not pass in the correct `ds_tp`, and the logic is a bit convoluted?

Thanks for pointing this out! I just fixed this.

I'm not sure how we should make this less convoluted, unless we do some refactoring work here. Any suggestions are welcome!

PR #1015, opened 2025-05-15
Support Async RLHF and Agent RLHF

xiaoxigua999:


iseesaw:
I want to ask if there is a specific comparison of training speed? Especially for larger [OPENRLHF_ASYNC_QUEUE_SIZE](https://github.com/OpenRLHF/OpenRLHF/blob/296db7ca515528969732066bbf6cb78f79aa494e/openrlhf/trainer/ppo_trainer_async.py#L315), how do training efficiency and performance compare? For example, comparisons similar to [AReaL](https://github.com/inclusionAI/AReaL).

hijkzzz:
> I want to ask if there is a specific comparison of training speed? Especially for larger [OPENRLHF_ASYNC_QUEUE_SIZE](https://github.com/OpenRLHF/OpenRLHF/blob/296db7ca515528969732066bbf6cb78f79aa494e/openrlhf/trainer/ppo_trainer_async.py#L315), how do training efficiency and performance compare? For example, comparisons similar to [AReaL](https://github.com/inclusionAI/AReaL).

There are two aspects here:

1. The async RL in AReal only supports FP16 training.
2. A queue size of 1 is the "most" stable option for off-policy training.

PR #1075, opened 2025-07-09
Enable logits extraction from vLLM for training

MooMoo-Yang:
This PR adds support for directly extracting logits from vLLM during training.
By enabling this feature, we avoid redundant computation of sequence logits during reinforcement learning (RL) fine-tuning, which previously required separate forward passes.

The change introduces a new flag:

```
--use_vllm_logprobs
```

When enabled, the training engine will use logits returned by vLLM instead of computing them again inside the trainer.

### DIFF
--- a/examples/scripts/train_reinforce_baseline_llama_ray_async_vllm_logits.sh
+++ b/examples/scripts/train_reinforce_baseline_llama_ray_async_vllm_logits.sh
@@ -1,0 +1,49 @@
+set -x
+
+# reinforce++
+
+ray job submit --address="http://127.0.0.1:8265" \
+   --runtime-env-json='{"working_dir": "/openrlhf"}' \
+   -- python3 -m openrlhf.cli.train_ppo_ray \
+   --ref_num_nodes 1 \
+   --ref_num_gpus_per_node 1 \
+   --reward_num_nodes 1 \
+   --reward_num_gpus_per_node 1 \
+   --actor_num_nodes 1 \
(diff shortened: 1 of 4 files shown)


MooMoo-Yang:
To evaluate the effectiveness of this change, we conducted four comparative experiments:

1. **Synchronous training** using the original training engine's logits.
2. **Asynchronous training** using the original training engine's logits.
3. Training using **vLLM-provided logits with BF16 precision**.
4. Training using **vLLM-provided logits with FP32 precision**, by modifying the model's `lm_head` to FP32 (to reduce numerical errors).

---

### ✅ Experimental Results
comparison



We found that using vLLM's **FP32 logits** achieves training performance comparable to that of the original engine, although a small performance gap still exists.

hijkzzz:
The async LLM for the agentic RL also needs to support vLLM logits.

examples/scripts/train_reinforce_baseline_llama_ray_async_vllm_logits.sh
```diff
@@ -0,0 +1,49 @@
+set -x
```
hijkzzz: remove this script~~

MooMoo-Yang:
> The async LLM for the agentic RL also needs to support vLLM logits.

Updated the corresponding training code. Local debugging shows no errors currently. Please review when you have time.

hijkzzz:
It needs to wait until the FP32 lm_head MR of vLLM is merged.

PR #1090, opened 2025-07-17
Fix remote rm reward format processing bug in experience_maker.py

LovelyQuantum:
The update_samples_with_rewards() function was incorrectly using torch.cat() directly on reward data from HTTP JSON requests. When rewards are passed via HTTP in JSON format, they come as Python lists rather than torch tensors, causing the function to fail.

The changes have been tested in my local environment.


hijkzzz:
this bug should be also fixed in make_experience ?

sherlockl01:
Yes, I also noticed the problem. What I'm even more curious about is, hasn't anyone else used this interface without finding any problems?

PR #1091, opened 2025-07-18
Support soft overlong punishment from DAPO

xjli360:
This PR adds the **overlong response penalty feature**, completing the implementation of **full DAPO support.**

### Key Differences from GRPO

Compared to GRPO, DAPO includes the following four main modifications:

✅ **Clip-Higher Strategy**: Already supported via the `--eps_clip_low_high` parameter, enabling asymmetric clipping.
✅ **Dynamic Filtering**: Enabled through `--dynamic_filtering` and `--dynamic_filtering_reward_range`, allowing adaptive sample filtering based on reward distribution.
✅ **Token-Level Loss**: Already supported; `PolicyLoss` uses `token_level_loss = True` by default for finer-grained optimization.
⭐️ **Soft Overlong Punishment**: Newly supported in this PR via `--overlong_buffer_len` and `--overlong_penalty_factor`.

### Overlong Response Penalty

To address the issue of excessively long responses, we introduce a **length-based penalty** that is smoothly applied based on how far the response exceeds a cache-safe length threshold. Specifically, if the response length \$|y|\$ exceeds a given safety buffer, the reward is adjusted as:

$$
R_{\text{length}}(y) =
\begin{cases}
0, & |y| \leq L_{\text{max}} - L_{\text{cache}} \\\\
\frac{(L_{\text{max}} - L_{\text{cache}}) - |y|}{L_{\text{cache}}}, & L_{\text{max}} - L_{\text{cache}} < |y| \leq L_{\text{max}} \\\\
-1, & |y| > L_{\text{max}}
\end{cases}
$$

This penalty is **added to the accuracy-based reward** during training to create a more balanced and stable learning signal. It discourages overly verbose outputs, improves training stability, and enhances overall model performance.

Our implementation is inspired by [verl/recipe/dapo](https://github.com/volcengine/verl/tree/main/recipe/dapo).

### DIFF
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -675,7 +675,8 @@ class RemoteExperienceMaker(ABC):
                 batch_size = len(response_lengths)
                 for j in range(batch_size):
                     valid_response_length = response_lengths[j].item()
-                    exceed_len = valid_response_length - expected_len
+                    # Cap the exceed_len to overlong_buffer_len to prevent excessive penalty
+                    exceed_len = min(valid_response_length - expected_len, overlong_buffer_len)
                     if exceed_len > 0:
                         overlong_penalty = -exceed_len / overlong_buffer_len * overlong_penalty_factor
                         # Apply penalty to the j-th reward in this experience
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -670,8 +670,9 @@ class RemoteExperienceMaker(ABC):
 
         # DAPO reward shaping with optional overlong penalty
         if args.overlong_buffer_len is not None:
-            assert args.generate_max_len >= args.overlong_buffer_len, (
-                "max_resp_len must be larger than overlong_buffer_len")
+            assert (
+                args.generate_max_len >= args.overlong_buffer_len
+            ), "max_resp_len must be larger than overlong_buffer_len"
             overlong_buffer_len = args.overlong_buffer_len
             expected_len = args.generate_max_len - overlong_buffer_len
             overlong_penalty_factor = args.overlong_penalty_factor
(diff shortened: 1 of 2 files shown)


hijkzzz:
Thanks! Will review it.

openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -668,6 +668,28 @@ def compute_advantages_and_returns(
...
+            ), "generate_max_len must be larger than overlong_buffer_len"
+            overlong_buffer_len = args.overlong_buffer_len
+            expected_len = args.generate_max_len - overlong_buffer_len
+            overlong_penalty_factor = args.overlong_penalty_factor
+            # Apply penalty to each reward based on response length
+            reward_idx = 0
+            for experience in experiences:
+                response_lengths = experience.info["response_length"]
+                batch_size = len(response_lengths)
+                for j in range(batch_size):
```
hijkzzz, line 676: I think there are errors here:

It is not compatible with dynamic batch.

Are the lengths of reward and response the same?
I suggest performing this transformation before the dynamic indices.

openrlhf/trainer/ppo_utils/experience_maker.py
MooMoo-Yang, line 1: The calculation can be performed between lines 599-611, which would be more appropriate as it ensures synchronous modification of each sample and its corresponding reward.

openrlhf/trainer/ppo_utils/experience_maker.py
xjli360: Thanks for the feedback!  I’ll verify and adapt the implementation to support dynamic batch in the next few days.

openrlhf/trainer/ppo_utils/experience_maker.py
```diff
@@ -660,6 +660,27 @@ def compute_advantages_and_returns(
...
+            overlong_penalty_factor = args.overlong_penalty_factor
+
+            # Apply penalty to each experience's rewards based on response length
+            for experience in experiences:
+                response_lengths = experience.info["response_length"]
+                batch_size = len(response_lengths)
+                for j in range(batch_size):
+                    valid_response_length = response_lengths[j].item()
+                    exceed_len = valid_response_length - expected_len
+                    if exceed_len > 0:
```
hijkzzz, line 680: when {y} > L_max the length reward should be -1 ?

xjli360: Yes, I capped `exceed_len` to `overlong_buffer_len` to avoid excessive penalty. Please review the code when you have time.

PR #1154, opened 2025-11-26
Default overlap_comm on for ZeRO-2+ RLHF runs

MagellaX:
Summary:
Flipped --overlap_comm to default ON across RLHF CLIs (PPO, SFT, DPO, KTO, KD, RM, PRM) using BooleanOptionalAction, so users can disable with --no-overlap_comm.
Guarded DeepSpeed config to set overlap_comm only for ZeRO stage ≥ 2; unchanged for ZeRO-0/1.
Tweaked README note to reflect the new default.

### DIFF
--- a/openrlhf/cli/train_dpo.py
+++ b/openrlhf/cli/train_dpo.py
@@ -5,6 +5,7 @@ from datetime import datetime
 
 from transformers.trainer import get_scheduler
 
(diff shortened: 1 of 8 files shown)


LYMDLUT:
I am not sure overlap_comm is a good choice, it may cause unstable loss spikes when the number of GPUs is large. I hope @xiaoxigua999 to further decide it.

PR #1156, opened 2025-12-07
add log logprobs_diff

richardodliu:
Pr to support print logprobs difference between actor and reference models.

### DIFF
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -636,7 +636,7 @@ class RemoteExperienceMaker(ABC):
                     base_action_log_probs,
                     kl_estimator=self.strategy.args.kl_estimator,
                 )
-                logprobs_diff = (action_log_probs.float() - base_action_log_probs.float())
+                logprobs_diff = action_log_probs.float() - base_action_log_probs.float()
             else:
                 kl = torch.zeros_like(action_log_probs, dtype=action_log_probs.dtype, device=device)
                 logprobs_diff = torch.zeros_like(action_log_probs, dtype=action_log_probs.dtype, device=device)


openrlhf/trainer/ray/ppo_actor.py
```diff
@@ -252,10 +256,14 @@ def training_step(self, experience: Experience, kl_ctl: float, step: int) -> Dic
...
+                logprobs_diff = action_log_probs.float() - base_action_log_probs.float()
             else:
-                kl = torch.zeros_like(action_log_probs, dtype=action_log_probs.dtype, device=action_log_probs.device)
+                kl = torch.zeros_like(action_log_probs)
+                logprobs_diff = torch.zeros_like(action_log_probs)
             kl_loss = masked_mean(kl, experience.action_mask)
+            logprobs_diff = masked_mean(logprobs_diff, experience.action_mask)
             experience.info["kl"] = kl_loss.detach()
+            experience.info["logprobs_diff"] = logprobs_diff.detach()
         else:
```
LYMDLUT: 303-308需要修改 training_step 方法，使用sum而不是mean来聚合样本级别的KL和长度

        # merge logs from info field
        # Extract response_length for weighted averaging
        response_length_tensor = None
        if "response_length" in experience.info:
            v = experience.info["response_length"]
            if isinstance(v, list):
                response_length_tensor = torch.tensor(v, dtype=torch.float)
            elif isinstance(v, torch.Tensor):
                response_length_tensor = v.float()

        for k, v in experience.info.items():
            if isinstance(v, list):
                v_tensor = torch.tensor(v, dtype=torch.float)
            elif isinstance(v, torch.Tensor):
                v_tensor = v.float()
            else:
                continue

            # For KL and logprobs_diff, compute weighted sum by response_length
            if k in ["kl", "logprobs_diff"] and response_length_tensor is not None:
                status[k] = (v_tensor * response_length_tensor).sum().item()
            # For length fields, compute sum instead of mean
            elif k in ["response_length", "total_length"]:
                status[k] = v_tensor.sum().item()
            # For other fields, use mean as before
            else:
                status[k] = v_tensor.mean().item()

        return status

hijkzzz:
why don't we use kl to measure this metric?

LYMDLUT:
> why don't we use kl to measure this metric?

as we may use k1&k2&k3 loss, these metrics are not consistent, we need a unified and most representative metric

PR #1160, opened 2025-12-17
fixes the numerical overflow risk of the k3 estimator in the compute_approx_kl function

CLFutureX:
**Background:** 
In the k3 calculation logic, if the input value of the log_ratio.exp() step is too large (e.g., 100), it will generate inf (infinity) due to exceeding the floating-point representation range. Although the final clamp can truncate the result, the intermediate inf may cause hidden issues such as gradient anomalies and dirty numerical data;

**Modify:** Add log_ratio.clamp(min=-10, max=10) before the exp() calculation in the k3 branch to limit the input range of the exponential function in advance, fundamentally preventing exp() from generating inf;

### DIFF
--- a/openrlhf/models/utils.py
+++ b/openrlhf/models/utils.py
@@ -35,6 +35,7 @@ def compute_approx_kl(
     if kl_estimator == "k3":
         log_ratio = log_probs.float() - log_probs_base.float()
         log_ratio = -log_ratio
+        log_ratio = log_ratio.clamp(min=-10, max=10)
         log_ratio = log_ratio.exp() - 1 - log_ratio
 
     log_ratio = log_ratio.clamp(min=-10, max=10)


CLFutureX:
@zhuzilin  hey， PTAL， thanks

CLFutureX:
@xiaoxigua999 @hijkzzz  hey, PTAL, Thanks

hijkzzz:
https://github.com/OpenRLHF/OpenRLHF/blob/591b148cc5a829c710078553cc490f178bb4a4af/openrlhf/models/utils.py#L40?

CLFutureX:
> https://github.com/OpenRLHF/OpenRLHF/blob/591b148cc5a829c710078553cc490f178bb4a4af/openrlhf/models/utils.py#L40
> 
> ?

Although there is a clamp operation afterward, if an overflow occurs in the earlier calculations, it will still affect the backpropagation process.

CLFutureX:
> https://github.com/OpenRLHF/OpenRLHF/blob/591b148cc5a829c710078553cc490f178bb4a4af/openrlhf/models/utils.py#L40
> 
> ?

K3 calculation leads to gradient anomalies due to numerical overflow
```python

def main():
    # ====================== 1. Simulate real-scenario inputs (trigger overflow) ======================
    # Construct log_probs and log_probs_base such that r = log_probs - log_probs_base = -100 
    # (far below 0, triggering numerical overflow in K3 calculation)
    log_probs = torch.tensor([-105.0], requires_grad=True)  # Log probabilities from model output
    log_probs_base = torch.tensor([-5.0], requires_grad=True)  # Log probabilities from base distribution
    r_original = log_probs - log_probs_base
    print(f"Original r value: {r_original.item()}")  # Output: -100.0

    # ====================== 2. Original K3 calculation (no clamping, triggers overflow) ======================
    def compute_kl_k3_original(log_probs, log_probs_base):
        r = log_probs.float() - log_probs_base.float()
        neg_r = -r
        # Original K3 formula (without expm1 for numerical stability)
        log_ratio = torch.exp(neg_r) - 1 - neg_r  
        # Post-clamping (cannot fix inf/nan caused by overflow)
        log_ratio = log_ratio.clamp(min=-10, max=10)
        return log_ratio

    # Forward calculation: exp(100) overflows to inf -> clamp still keeps inf
    kl_original = compute_kl_k3_original(log_probs, log_probs_base)
    print(f"Original K3 forward result: {kl_original.item()}")  # Output: inf (overflow)

    # Backpropagation: inf leads to nan gradient (IEEE 754 floating-point rule)
    loss_original = kl_original.sum()
    loss_original.backward()
    print(f"log_probs gradient (original K3): {log_probs.grad.item()}")  # Output: nan
    print(f"log_probs_base gradient (original K3): {log_probs_base.grad.item()}")  # Output: nan
```

LYMDLUT:
For kl loss，we suggest all of you to use k2 rather than k3. Actually, k3 should never be used

PR #1152, opened 2025-11-21
feat: restructure and clean up streaming async sampling (#1130)

Freder-chen:

PR #1163, opened 2025-12-23
Fix SignalActor concurrency issues and refactor state management based on condition lock

CLFutureX:
[ISSUE](https://github.com/OpenRLHF/OpenRLHF/issues/1164)

### Problem Background
The current `SignalActor` uses `asyncio.Event` to implement mutual exclusion control between trajectory generation and weight update. In the OpenRLHF distributed training scenario, it has the following core issues:
1. **Mutual Exclusion Logic Failure**: Two independent `Event` objects cannot guarantee operation mutual exclusion. Trajectory generation and weight update may execute simultaneously, leading to model weight dirty writes and abnormal sample generation.
2. **Concurrent Call Risks**: Unlocked state modifications may be tampered with by concurrent requests, making the execution results of `set_*` methods uncontrollable.

### Solution
Refactor `SignalActor` based on `asyncio.Condition` and enum states to achieve atomic state management and strict mutual exclusion:
1. **Introduce State Enum**: Define three states (`IDLE/GENERATOR/UPDATE_WEIGHT`) to clearly identify the current resource occupant.
2. **Condition Lock Atomic Operations**: All state modification and waiting logic are executed in `async with self.cond` to ensure concurrency safety.
3. **Loop Wait to Prevent Spurious Wakeups**: Use `while` loop instead of `if` to judge the state, solving the spurious wakeup problem of condition locks.
4. **Compatible with Legacy Interfaces**: Retain `set_generating/set_update_weights` methods, enabling seamless migration without modifying business code.

### Core Changes
- Add `SignalStatus` enum class to standardize state definitions.
- Replace `asyncio.Event` with `asyncio.Condition` to implement atomic state modification and notification mechanism.
- Add `release_*` methods to support manual state release; supplement the internal `_force_release_and_block` method to strengthen mutual exclusion logic.
- Fix the syntax error where `state` was not prefixed with `self` in the original code.

### DIFF
--- a/openrlhf/trainer/ppo_trainer_async.py
+++ b/openrlhf/trainer/ppo_trainer_async.py
@@ -19,7 +19,7 @@ logger = init_logger(__name__)
 
 class SignalStatus(Enum):
     IDLE = 0  # Idle state (can be occupied by either operation)
-    GENERATOR = 1  # Occupied by trajectory generation
+    GENERATION = 1  # Occupied by trajectory generation
     UPDATE_WEIGHT = 2  # Occupied by weight update
 
 
@@ -38,12 +38,13 @@ class SignalActor:
         """
         async with self.cond:
             # Loop to avoid spurious wakeups: wait until weight update is finished
-            while self.state == SignalStatus.UPDATE_WEIGHT:
+            while self.state != SignalStatus.IDLE:
+                if self.state == SignalStatus.GENERATION:
+                    return
                 await self.cond.wait()
 
-            # Atomically occupy state for generation if idle
-            if self.state == SignalStatus.IDLE:
-                self.state = SignalStatus.GENERATOR
+            # Atomically occupy state for generation
+            self.state = SignalStatus.GENERATION
 
     async def wait_update_weights(self):
         """Wait for weight update to be allowed.
@@ -52,15 +53,16 @@ class SignalActor:
         """
         async with self.cond:
             # Loop to avoid spurious wakeups: wait until generation is finished
-            while self.state == SignalStatus.GENERATOR:
+            while self.state != SignalStatus.IDLE:
+                if self.state == SignalStatus.UPDATE_WEIGHT:
+                    return
                 await self.cond.wait()
 
-            # Atomically occupy state for weight update if idle
-            if self.state == SignalStatus.IDLE:
-                self.state = SignalStatus.UPDATE_WEIGHT
+            # Atomically occupy state for weight update
+            self.state = SignalStatus.UPDATE_WEIGHT
 
     def release_generating(self):
-        asyncio.create_task(self._release_state(SignalStatus.GENERATOR))
+        asyncio.create_task(self._release_state(SignalStatus.GENERATION))
 
     def release_update_weights(self):
         asyncio.create_task(self._release_state(SignalStatus.UPDATE_WEIGHT))


CLFutureX:
@xiaoxigua999  hey, PTAL， Thanks

Freder-chen:
This was already fixed in PR #1152 — specifically in: https://github.com/OpenRLHF/OpenRLHF/pull/1152/commits/341bd9a12260116e2231130f6237ed02c699c4a9

Could you please double-check that this change addresses the issue in this PR #1163 as well?

CLFutureX:
> This was already fixed in PR #1152 — specifically in: [341bd9a](https://github.com/OpenRLHF/OpenRLHF/commit/341bd9a12260116e2231130f6237ed02c699c4a9)
> 
> Could you please double-check that this change addresses the issue in this PR #1163 as well?

Oops, I didn't notice this earlier. I checked this commit [341bd9a](https://github.com/OpenRLHF/OpenRLHF/commit/341bd9a12260116e2231130f6237ed02c699c4a9) found there's still an issue in some concurrent scenarios. For example, let's say there are two requests to update weights: after the first request finishes execution, it releases the lock and sets _updating =False. Meanwhile, the second weight update request is still in progress. If a generation request comes in at this moment, it will also acquire the lock, which causes concurrency issues.

Freder-chen:
> Oops, I didn't notice this earlier. I checked this commit [341bd9a](https://github.com/OpenRLHF/OpenRLHF/commit/341bd9a12260116e2231130f6237ed02c699c4a9) found there's still an issue in some concurrent scenarios. For example, let's say there are two requests to update weights: after the first request finishes execution, it releases the lock and sets _updating =False. Meanwhile, the second weight update request is still in progress. If a generation request comes in at this moment, it will also acquire the lock, which causes concurrency issues.

Thanks for the review. In the current implementation, acquire_for_update() only waits on _generating, so two update callers could enter concurrently, and the first release_for_update() would set _updating = False while the second update is still running. In our current setup, updates are issued by a single TrainingActor and are effectively serialized (single rollout + single train), so we don’t expect concurrent updates today. That said, I agree the lock should be correct by construction: I’ll make updates writer-exclusive by also gating on _updating (e.g., waiting while _updating is True, or adding an assert) so generation can’t proceed until the only in-flight update has finished.

Also, PR #1152 really needs reviewers right now—would you be able to take a look?

Freder-chen:
> > This was already fixed in PR #1152 — specifically in: [341bd9a](https://github.com/OpenRLHF/OpenRLHF/commit/341bd9a12260116e2231130f6237ed02c699c4a9)
> > Could you please double-check that this change addresses the issue in this PR #1163 as well?
> 
> Oops, I didn't notice this earlier. I checked this commit [341bd9a](https://github.com/OpenRLHF/OpenRLHF/commit/341bd9a12260116e2231130f6237ed02c699c4a9) found there's still an issue in some concurrent scenarios. For example, let's say there are two requests to update weights: after the first request finishes execution, it releases the lock and sets _updating =False. Meanwhile, the second weight update request is still in progress. If a generation request comes in at this moment, it will also acquire the lock, which causes concurrency issues.

Fixed by introducing a dedicated Ray lock actor (asyncio.Lock) to guard the vLLM critical section, providing cross-actor mutual exclusion via await acquire() / release(). I think this is the simplest and most readable approach.

Reference: https://github.com/OpenRLHF/OpenRLHF/blob/ece5602fb415a131126185935183d8e35439303e/openrlhf/trainer/ppo_trainer_async.py#L18-L28

hijkzzz:
Thanks for you MR!!!!

Issue #1128, opened 2025-09-04
Token and text concatenation issue in the process of agentic training.


qianxiao1111:
After migrating from version 0.8.5 to 0.8.10, the training performance significantly deteriorated when using the same code.

1）The reward increases more slowly;

2）The post-training checkpoint not only fails to improve across various benchmarks but actually performs worse.

After carefully inspecting the code in agent.py (in version 0.8.5 it was in vllm_engine_async.py), I found that there might be some issues with the current token and text concatenation.

`                request_output = await self.generate(current_obs_tokens, sampling_params)
                action_tokens = request_output.outputs[0].token_ids
                action_text = request_output.outputs[0].text

                # Record action range in token space
                action_start = len(current_obs_tokens)
                action_end = action_start + len(action_tokens)
                action_ranges.append((action_start, action_end))

                # Call step function to get environment feedback
                states = {
                    "observation_text": observation_text,
                    "action_text": action_text,
                    "label": label,
                    "sampling_params": sampling_params,
                }
                step_result = await agent_instance.step.remote(states)

                total_reward += step_result["rewards"].item()
                final_scores = step_result.get("scores", total_reward)
                environment_feedback_text = step_result["environment_feedback"]
                done = step_result["done"]
                extra_logs = step_result.get("extra_logs", {})

                # Concatenate observation, action, and environment_feedback, then tokenize
                observation_text = observation_text + action_text + environment_feedback_text
                current_obs_tokens = (
                    current_obs_tokens
                    + action_tokens
                    + self.hf_tokenizer(environment_feedback_text, add_special_tokens=False, return_tensors="pt")[
                        "input_ids"
                    ][0].tolist()
                )`
There may be two problem:
1)It seems that the output action_text does not include the eos_token and requires manual concatenation. Accordingly, the action_text may also need to be concatenated with the eos_token_id.
2)In env_feed_back, it is necessary to manually concatenate the special_token, and a newline character "\n" must be added before the special_token.

For issue (1), it seems that the following code should be added in the agent.py file:
`                if action_tokens[-1] != self.hf_tokenizer.eos_token_id:
                    action_tokens.append(self.hf_tokenizer.eos_token_id)   `

For issue (2), it may be necessary to manually modify both the action and the environment_feedback_text in the agent_func.py file.

However, I cannot confirm whether the above modifications are entirely correct. Kindly request @xiaoxigua to review them.

hijkzzz:
[...]

PR #1181, opened 2026-01-30
feat: add AgentTokenHandler with defensive token concatenation for agentic training (#1128)

ichbinlucaskim:
## Problem
GitHub issue #1128 reports significant performance degradation in agentic RLHF 
training due to:
1. Missing EOS tokens in action sequences
2. Misaligned turn boundaries in multi-turn training
3. No validation mechanism for token offset integrity

## Root Cause
vLLM's generated tokens may not include EOS, and environment feedback needs 
explicit newline markers for proper turn-boundary recognition.

## Solution: Strategy Pattern + Defensive Programming
This PR introduces an **abstract interface** that allows:
- Model-specific tokenization strategies via subclassing
- Turn-boundary integrity validation with Defensive Programming
- Observability through ConcatenationReport
- Framework remains agnostic to token quirks

## Implementation
- `AgentTokenHandler` (abstract base class)
- `DefaultAgentTokenHandler` (sensible defaults with null-safety)
- `ConcatenationReport` (validation & observability)
- Example: `agent_func_with_safe_tokenization.py`

### Defensive Programming Measures
- **Zero-Inference Filtering**: Removes None/empty sequences before processing
- **State Collision Guard**: Detects duplicate source keys with warnings
- **Mathematical Precision**: Maintains offset calculation invariant
- **Strict Type Hinting**: All variables explicitly annotated

## Testing
✓ Turn-boundary integrity validated
✓ Token concatenation: observation(0:14) + action(14:18) + feedback(18:28)
✓ ConcatenationReport validation: valid=True, warnings=0
✓ Graceful degradation on None/invalid inputs (no crashes)

## Usage
```python
from openrlhf.utils.agent_tokenization import DefaultAgentTokenHandler

token_handler = DefaultAgentTokenHandler(tokenizer, verbose=True)
action_seq = token_handler.handle_action_tokens(action_tokens, action_text)
feedback_seq = token_handler.handle_feedback_tokens(feedback_text)
concatenated, report = token_handler.concatenate_sequences([action_seq, feedback_seq])

assert report.is_valid  # Turn boundaries align perfectly


hijkzzz:
@Freder-chen

Freder-chen:
I believe this issue no longer exists in the latest version. Both `current_obs_tokens` and `action_tokens` already include the correct special tokens (depending on the tokenizer configuration). As I understand it, `environment_feedback_text` expects the environment to embed any required special tokens; the current implementation is intentionally simple and easy to follow.

Given that, I’m not sure what problem this pull request is trying to solve. Could you clarify the concrete failure case or expected behavior change?

cite: https://github.com/OpenRLHF/OpenRLHF/blob/main/openrlhf/utils/agent.py#L109-L110

ichbinlucaskim:
Thanks for the clarification and pointer to the current implementation.

Just to clarify my motivation: after reading the issues described in #1128 about token/text concatenation, I felt that a small adjustment in this area could still be helpful for users, especially in edge cases where similar problems may persist. Those issues also reminded me of similar edge cases I’ve encountered in other libraries when stitching together multi‑turn sequences.

In those situations, having a small, opt‑in utility that

- makes EOS / turn‑boundary handling explicit
- validates token offsets and reports suspicious patterns, and
- has been very helpful for users when debugging subtle training regressions.

That’s why I proposed AgentTokenHandler: not to change the default behavior or override the framework’s design, but to give users a reusable, defensive helper they can enable when they suspect token‑boundary issues.

If you feel this is better positioned as a utility or example (rather than in the core agent path), I’m happy to

- move it under utils or examples, and
- adjust the API/naming so it’s clearly an optional debugging tool.

Also, to be clear: when the comments/docstrings say this handler “addresses issue #1128”, it should be read as “inspired by / designed to help debug issues of that type”, not as a claim that there is a currently reproducible bug in the latest main branch. If any of my wording around issue #1128 came across as misleading or presumptive, that wasn’t my intention, and I’m happy to rephrase or adjust it.

Please let me know what placement and scope would fit best with OpenRLHF’s design philosophy, and I can update the PR accordingly.

I really appreciate you taking the time to look at this and I’m happy to adjust or close the PR if it doesn’t fit the project’s direction.

PR #1186, opened 2026-02-10
Support transformers v5: update deprecated APIs and dependencies 

jiosephlee:
## Summary                                                                                                         
  - Replace `batch_decode()` with `decode()` — in transformers v5, `batch_decode` and `decode` are unified into a    
  single `decode` method that handles both single and batch inputs                                                   
  - Update `transformers` pin from `==4.57.6` to `>=5.0.0`                                                           
  - Add `huggingface_hub>=1.0.0` dependency (required by transformers v5)                                            

  See the [transformers v5 migration guide](https://huggingface.co/docs/transformers/main/en/v5_migration_guide) for full details.

## Issues
- There might be breaking changes with transformers v5 regarding how it interacts with other libraries that are still migrating. There is an incompatibility with apply_chat_template(..., tokenize=True), for instance.
- I've had some challenges with DeepSpeed using Transformers v5-specific models as well, but those issues should be confined to those models. 
- I'm a bit new to the space, so there might be other issues I haven't caught or taken into consideration.

### DIFF
--- a/openrlhf/cli/batch_inference.py
+++ b/openrlhf/cli/batch_inference.py
@@ -173,7 +173,11 @@ def batch_generate(args):
                 pad_token_id=tokenizer.pad_token_id,
                 eos_token_id=tokenizer.eos_token_id,
             )
-            outputs = tokenizer.decode(outputs, skip_special_tokens=True) if _TRANSFORMERS_V5 else tokenizer.batch_decode(outputs, skip_special_tokens=True)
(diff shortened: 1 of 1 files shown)


jiosephlee:
It might be better to push this off for later? Some previous models are not compatible with v5 and will throw errors like
`AttributeError: InternS1Tokenizer has no attribute _added_tokens_decoder`

PR #1199, opened 2026-03-19
fix: wrap grad_norm with float() for Ray serialization in colocate mode

yxs:
DeepSpeed's `get_global_grad_norm()` may return a non-plain-float value that fails Ray's pickle serialization when sent from GPU actor to CPU PPOTrainer in colocate mode. This causes:

  RuntimeError: Attempting to deserialize object on a CUDA device
  but `torch.cuda.is_available()` is False.

Wrapping with float() ensures the value is a plain Python float before being serialized by Ray.

Fixes the grad_norm logging added in 4908e6e.

Verified on 2x H100 with Qwen2.5-1.5B-Instruct + GSM8K (PPO training)

image

image


hijkzzz:
fixed

PR #1228, opened 2026-04-26
Add SFT tools field support for chat templates

### LINKED ISSUE #1079: SFT dataset 建议支持tools字段
目前sft dataset只接受了messages，如下面代码中的data[input_key]，建议也对tools字段进行支持，即data存在tools字段的话，`apply_chat_template(chat, data['tools'], xxx)`

```python
def preprocess_data(
    data, input_template=None, input_key="input", output_key=None, apply_chat_template=None, multiturn=False
):
    if apply_chat_template:
        if output_key:
            prompt_message = data[input_key]
            response_message = data[output_key]

            if isinstance(prompt_message, str) and isinstance(response_message, str):
                prompt_message = [{"role": "user", "content": prompt_message}]
                response_message = [{"role": "assistant", "content": response_message}]

            prompt = apply_chat_template(prompt_message, tokenize=False, add_generation_prompt=True)
            response = apply_chat_template(prompt_message + response_message, tokenize=False)[len(prompt) :]
        else:
            prompt = apply_chat_template(data[input_key][:-1], tokenize=False, add_generation_prompt=True)
            response = apply_chat_template(data[input_key], tokenize=False)[len(prompt) :]
    else:
        prompt = data[input_key]
        if input_template:
            prompt = input_template.format(prompt)
        # output_key is None for continue pretrain
        response = data[output_key] if output_key else ""
    return prompt, response
```

taivu1998:
## Summary

- Forward an optional row-level `tools` field into SFT chat-template rendering when `--data.apply_chat_template` is enabled.
- Apply the same forwarding in split `input`/`output`, full-conversation, and multiturn response-range preprocessing paths.
- Document the supported `tools` field and add focused tests for tool-calling SFT dataset behavior.

## Motivation

Fixes #1079.

SFT preprocessing previously passed only the conversation messages to `apply_chat_template(...)`. Tool-calling chat templates in HuggingFace tokenizers can also use `tools=...`, but OpenRLHF dropped that row-level dataset field before rendering. As a result, datasets that already contained tool schemas could not render tool-aware SFT prompts through the built-in chat-template path.

## Design

- Add a small helper that returns `{"tools": data["tools"]}` only when the row contains a non-`None` `tools` value.
- Preserve existing behavior for rows without `tools` or with `tools=None`.
- Preserve empty tool lists by forwarding `tools=[]`, which lets the tokenizer/template decide how to render an explicitly empty tool set.
- Keep the change scoped to SFT preprocessing, matching the issue request without adding a new CLI flag or changing reward/prompt dataset behavior.

## Validation

- `python3.10 -m py_compile openrlhf/datasets/sft_dataset.py tests/test_sft_dataset_tools.py`
- `python3.10 -m pytest tests/test_sft_dataset_tools.py -q`
- `python3.10 -m pytest tests/test_ray_env_vars.py tests/test_sft_dataset_tools.py -q`
- `git diff --check`
- Repo pre-commit hooks on commit: `autoflake`, import formatting, and Black formatting passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,8 @@ OpenRLHF provides flexible data processing methods:
 - `--data.prompt_probs` / `--data.dataset_probs`: Mix multiple datasets (e.g., `0.1,0.4,0.5`)
 - `--eval.dataset`: Specify evaluation dataset path
 
+For SFT datasets with `--data.apply_chat_template`, each row may also include a `tools` field. OpenRLHF forwards it to HuggingFace chat templates so tool-calling templates can render the available tool schemas.
+
 **Chat Template Example**:
 
 ```python
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ OpenRLHF provides flexible data processing methods:
 - `--data.prompt_probs` / `--data.dataset_probs`: Mix multiple datasets (e.g., `0.1,0.4,0.5`)
 - `--eval.dataset`: Specify evaluation dataset path
 
-For SFT datasets with `--data.apply_chat_template`, each row may also include a `tools` field. OpenRLHF forwards it to HuggingFace chat templates so tool-calling templates can render the available tool schemas.
+For SFT datasets with `--data.apply_chat_template`, each row may also include a `tools` field (a list of dictionaries following the HuggingFace tool-use format). OpenRLHF forwards it to HuggingFace chat templates so tool-calling templates can render the available tool schemas.
 
 **Chat Template Example**:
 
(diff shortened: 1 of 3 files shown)


Excelius-Wang (commented):
I checked this with a local Transformers tokenizer and an Arrow-backed dataset, in addition to the seven added tests. In a user → assistant → tool → assistant conversation, the tool schema now reaches the rendered prompt, both assistant answers receive loss, and the schema plus tool response stay masked out. The same checks pass for `tools=None` and `tools=[]`; reverting the production change drops the populated schema.

Passing the same tools to the prefix and full-conversation renders is important for keeping those response offsets consistent. I didn't find a blocking issue in that path. This used a small local tool-aware template, not a downloaded model template or a training run.

PR #1240, opened 2026-05-17
fix: ignore masked invalid values in policy loss reductions

haoyang9804:
## Summary

This fixes a quiet PPO training-signal bug in `PolicyLoss.forward`.

`action_mask=0` positions can contain padding/tool/rejected-token values that are semantically ignored by the loss. Before this change, `PolicyLoss` still formed `log_probs - old_log_probs` and later used mask multiplication in reduction paths. If a masked `old_log_probs` slot was `NaN`, both actor loss and `ppo_kl` became `NaN` even though the token was masked out.

The fix zeros masked values before policy-ratio arithmetic and zeros masked per-token losses before `aggregate_loss` reductions. Unmasked invalid values are not hidden.

## Concrete Failure Shape

The smallest failing batch has one valid action token and one masked token:

```text
log_probs:     [[0.0, 0.0]]
old_log_probs: [[0.0, NaN]]
advantages:    [[1.0, 1.0]]
action_mask:   [[1.0, 0.0]]
```

The second token is semantically ignored because `action_mask=0`. It may come from padding, a filtered/rejected rollout position, or another non-trainable response slot. Before this fix, `PolicyLoss.forward` still computed:

```python
raw_policy_log_ratio = log_probs - old_log_probs
```

That produces:

```text
raw_policy_log_ratio: [[0.0, NaN]]
```

Later mask multiplication does not remove the invalid value:

```text
NaN * 0 = NaN
```

So a token that should not contribute to training can still make `loss` and `ppo_kl` non-finite. The fixed path zeroes masked values before ratio/KL/loss arithmetic, so the masked slot stays semantically ignored and the valid token keeps the same gradient.

## Reproduction Recipe

```json
{
  "bug_id": "BUG-4B85D84D7C0E",
  "target": "openrlhf",
  "boundary": "openrlhf.models.loss.PolicyLoss.forward",
  "scenario": {
    "log_probs": [[0.0, 0.0]],
    "old_log_probs": [[0.0, "nan"]],
    "advantages": [[1.0, 1.0]],
    "action_mask": [[1.0, 0.0]]
  },
  "expected_unpatched": {
    "loss_isfinite": false,
    "ppo_kl_isfinite": false
  },
  "expected_fixed": {
    "loss_isfinite": true,
    "ppo_kl_isfinite": true,
    "grad_isfinite": true
  }
}
```

## Portable Runner

```bash
#!/usr/bin/env bash
set -euo pipefail

REPO="${REPO:-$(pwd)}"
export PYTHONPATH="${REPO}${PYTHONPATH:+:${PYTHONPATH}}"

python3 - <<'PY'
import json
import math
import torch

from openrlhf.models.loss import PolicyLoss


def scalar(x):
    value = x.detach().cpu().item()
    if isinstance(value, float) and math.isnan(value):
        return "nan"
    return value


log_probs = torch.nn.Parameter(torch.tensor([[0.0, 0.0]], dtype=torch.float32))
old_log_probs = torch.tensor([[0.0, float("nan")]], dtype=torch.float32)
advantages = torch.tensor([[1.0, 1.0]], dtype=torch.float32)
action_mask = torch.tensor([[1.0, 0.0]], dtype=torch.float32)

loss, clip_ratio, ppo_kl, vllm_kl = PolicyLoss(policy_loss_type="ppo")(
    log_probs=log_probs,
    old_log_probs=old_log_probs,
    advantages=advantages,
    action_mask=action_mask,
)
loss.backward()

print(json.dumps({
    "loss": scalar(loss),
    "loss_isfinite": bool(torch.isfinite(loss).item()),
    "clip_ratio": scalar(clip_ratio),
    "clip_ratio_isfinite": bool(torch.isfinite(clip_ratio).item()),
    "ppo_kl": scalar(ppo_kl),
    "ppo_kl_isfinite": bool(torch.isfinite(ppo_kl).item()),
    "grad": log_probs.grad.detach().cpu().tolist(),
    "grad_isfinite": bool(torch.isfinite(log_probs.grad).all().item()),
    "vllm_kl": vllm_kl,
}, indent=2))
PY
```

## Observed Output

Unpatched OpenRLHF `main@c3188af37cec984614aaa38906e71fa2fc57b079`:

```json
{
  "loss": "nan",
  "loss_isfinite": false,
  "ppo_kl": "nan",
  "ppo_kl_isfinite": false,
  "grad_isfinite": true,
  "update_delta_isfinite": true
}
```

Fixed branch `b99673fa1547a9da8b0762fa5dbb0229eb672b7c`:

```json
{
  "loss": -1.0,
  "loss_isfinite": true,
  "ppo_kl": 0.0,
  "ppo_kl_isfinite": true,
  "grad_isfinite": true,
  "update_delta_isfinite": true
}
```

The fixed validation imports `openrlhf.models.loss` from the repaired checkout and starts the local Ray runtime; vLLM and DeepSpeed imports also pass.

## Duplicate Search

OpenRLHF has no PR template. I followed `CONTRIBUTING.md`, which asks contributors to use pre-commit hooks.

Searched OpenRLHF upstream PRs/issues for exact and related reports:

- `repo:OpenRLHF/OpenRLHF is:pr PolicyLoss old_log_probs NaN action_mask`: 0 results
- `repo:OpenRLHF/OpenRLHF is:pr ppo_kl NaN action_mask`: 0 results
- `repo:OpenRLHF/OpenRLHF is:issue PolicyLoss old_log_probs NaN action_mask`: 0 results

Same-family local RL-infra bugs exist in other projects, but they are not OpenRLHF `PolicyLoss.forward` fixes.

## Tests

```bash
python3 -m pytest tests/test_loss_aggregation.py -q
python3 -m ruff check openrlhf/models/loss.py tests/test_loss_aggregation.py
pre-commit run --files openrlhf/models/loss.py tests/test_loss_aggregation.py
```

Results:

- `tests/test_loss_aggregation.py`: 8 passed
- `ruff`: passed
- `pre-commit`: passed

### DIFF
--- a/openrlhf/models/loss.py
+++ b/openrlhf/models/loss.py
@@ -8,6 +8,12 @@ import torch.nn.functional as F
 from .utils import masked_mean
 
 
+def _zero_masked_values(tensor: Optional[torch.Tensor], mask: Optional[torch.Tensor]) -> Optional[torch.Tensor]:
+    if tensor is None or mask is None:
+        return tensor
+    return tensor.masked_fill(~mask.bool(), 0.0)
+
+
 def aggregate_loss(
     loss: torch.Tensor,
     loss_mask: torch.Tensor,
@@ -17,6 +23,8 @@ def aggregate_loss(
     global_batch_size: Optional[float] = None,
 ) -> torch.Tensor:
     """Aggregate token losses using verl-compatible normalization."""
+    loss = _zero_masked_values(loss, loss_mask)
+
     if token_level_loss:
         if batch_num_tokens is None:
             return masked_mean(loss, loss_mask, dim=None)
@@ -154,6 +162,14 @@ class PolicyLoss(nn.Module):
         batch_num_tokens: Optional[float] = None,
         global_batch_size: Optional[float] = None,
     ) -> torch.Tensor:
+        if action_mask is None:
+            action_mask = torch.ones_like(log_probs)
+
+        log_probs = _zero_masked_values(log_probs, action_mask)
+        old_log_probs = _zero_masked_values(old_log_probs, action_mask)
+        advantages = _zero_masked_values(advantages, action_mask)
+        rollout_log_probs = _zero_masked_values(rollout_log_probs, action_mask)
+
         raw_policy_log_ratio = log_probs - old_log_probs
         if self.policy_loss_type == "ppo":
             policy_log_ratio = raw_policy_log_ratio.clamp(min=-20.0, max=20.0)
--- a/tests/test_loss_aggregation.py
+++ b/tests/test_loss_aggregation.py
@@ -93,6 +93,18 @@ def test_loss_batch_info_excludes_fully_masked_sequences():
     assert torch.allclose(seq_loss, torch.tensor(2.0))
 
 
+def test_aggregate_loss_ignores_masked_nonfinite_values():
+    loss = torch.tensor([[1.0, float("nan")], [3.0, float("inf")]])
+    mask = torch.tensor([[1.0, 0.0], [1.0, 0.0]])
+
+    assert torch.allclose(aggregate_loss(loss, mask), torch.tensor(2.0))
+    assert torch.allclose(
+        aggregate_loss(loss, mask, dp_size=1, batch_num_tokens=mask.sum()),
+        torch.tensor(2.0),
+    )
+    assert torch.allclose(aggregate_loss(loss, mask, token_level_loss=False), torch.tensor(2.0))
+
+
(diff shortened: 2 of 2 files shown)


Excelius-Wang (commented):
Zeroing masked inputs before the ratio arithmetic addresses the part that multiplication by a mask cannot fix: `NaN * 0` is still NaN.

I checked masked NaNs in current log-probabilities, old log-probabilities and advantages for both PPO and GSPO. The loss and diagnostics match the valid-token-only case, and the masked current-log-probability gradient is zero. A NaN at an unmasked position still remains visible. The existing regressions and these additional CPU checks pass. I haven't run distributed training.

PR #1246, opened 2026-05-30
Fix overlong penalty action token length

### LINKED ISSUE #1243: overlong_penalty should exclude tool response tokens from length calculation (use action_mask/action_ranges)
when doing "agentic" training (i.e. training with tools), overlong buffer uses the `experience.response length` as per this following line:

https://github.com/OpenRLHF/OpenRLHF/blob/6c6056daa522e2216466f6e0351cbe453434e185/openrlhf/trainer/ppo_utils/length_penalty.py#L45

response length includes tokens from tool response as per following line:

https://github.com/OpenRLHF/OpenRLHF/blob/3a9815c51aa7e985983eacd06504aaf6db0bbe6b/openrlhf/trainer/ppo_utils/samples_generator.py#L277

given tool responses can't be optimised by model, this can add a bit of noise to "agentic" training and worse can discourage model to use tools.

we should compute response length using `action_mask` or `action_ranges` inside overlong penalty function.

Happy to create PR to fix this. Please let me know if you would like me to.

Jiang020609:
Fixes #1243. Uses action_mask when available for overlong penalty length so tool response tokens are excluded from trainable response length accounting. Tests: python -m pytest tests; python -m pre_commit run --files openrlhf/trainer/ppo_utils/length_penalty.py tests/test_length_penalty.py

### DIFF
--- a/openrlhf/trainer/ppo_utils/length_penalty.py
+++ b/openrlhf/trainer/ppo_utils/length_penalty.py
@@ -12,12 +12,27 @@ from openrlhf.utils.logging_utils import init_logger
 
 logger = init_logger(__name__)
 
+_warned_missing_action_mask = False
+
 
 def _get_overlong_response_lengths(experience):
+    global _warned_missing_action_mask
+
     action_mask = getattr(experience, "action_mask", None)
     if action_mask is not None:
         response_lengths = action_mask.sum(dim=-1)
     else:
+        # Falling back to response_length counts every generated token, including
+        # non-action tokens such as tool responses. This is the pre-fix behavior and
+        # is only correct for non-agentic runs. A missing action_mask during agentic
+        # training is most likely a misconfiguration, so surface it once.
+        if not _warned_missing_action_mask:
+            logger.warning(
+                "action_mask is not set on experiences; overlong penalty falls back to "
+                "response_length, which counts non-action tokens (e.g. tool responses). "
+                "If this is an agentic run, ensure action_mask is populated."
+            )
+            _warned_missing_action_mask = True
         response_lengths = getattr(experience, "response_length", None)
 
     if response_lengths is None:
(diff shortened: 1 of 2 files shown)


Jiang020609:
Addressed the review feedback in 78f8dd8: action_mask now uses bool tensor sum directly, and missing length metadata raises a clear ValueError. Added a unit test for the missing action_mask/response_length case.\n\nTests:\n- python -m pytest tests/test_length_penalty.py\n- python -m pytest tests\n- python -m pre_commit run --files openrlhf/trainer/ppo_utils/length_penalty.py tests/test_length_penalty.py

openrlhf/trainer/ppo_utils/length_penalty.py
```diff
@@ -13,6 +13,22 @@
 logger = init_logger(__name__)
 
 
+def _get_overlong_response_lengths(experience):
+    action_mask = getattr(experience, "action_mask", None)
+    if action_mask is not None:
+        response_lengths = action_mask.sum(dim=-1)
+    else:
+        response_lengths = getattr(experience, "response_length", None)
```
KamayaniR, line 36: the fallback here (getattr(experience, "response_length", None)) silently preserves the original buggy behavior when action_mask isn't set. In agentic training, a missing action_mask is likely a misconfiguration rather than an intentional choice, so this could mask the issue rather than surface it. Might be worth logging a warning in the fallback branch, or considering whether action_mask should be required for agentic runs.

Jiang020609: Good point, thanks. I've added a warning in the fallback branch (commit abba1bb): when `action_mask` is missing we now emit a `logger.warning` noting that `response_length` counts non-action tokens such as tool responses, and that agentic runs should make sure `action_mask` is populated. It's guarded to fire only once per process so it doesn't spam the per-experience loop.

I kept it as a warning rather than hard-requiring `action_mask`, since non-agentic RLHF runs legitimately have no `action_mask` and the `response_length` fallback is correct for them — making it mandatory would break those cases. The warning surfaces the likely misconfiguration without changing valid behavior. Also added a unit test (`test_overlong_penalty_warns_once_when_falling_back_without_action_mask`) covering the warn-once path.

PR #1247, opened 2026-06-05
Added REBEL-inspired offline reward-gap regression loss to DPO trainer

LeoPhilly:
## Summary

Adds an offline reward-gap regression loss inspired by [REBEL: Reinforcement Learning via Regressing Relative Rewards](https://arxiv.org/abs/2404.16767) as an opt-in option to the DPO trainer pipeline. Enabled via `--model.rebel_enable` and parameterized by `--model.eta`. Existing DPO and IPO code paths are unchanged when the flag is off.

Note this is not a paper-faithful implementation of the full REBEL algorithm. It is an offline, one-iteration proxy — a drop-in loss inside the existing DPO trainer, without policy iteration or online generation. The paper's theoretical guarantees (e.g. convergence) do not apply.

## What's added

- New `REBELLoss` class in `openrlhf/models/loss.py` (separate from DPO/IPO, since REBEL takes different inputs — a per-example `rewards_margin` and an `eta` parameter)
- `--model.rebel_enable` and `--model.eta` CLI flags, gated so existing DPO/IPO behavior is unaffected when off
- Margin column handling in the dataset path; REBEL expects a per-example margin column, while DPO/IPO ignore it.
- 6 unit tests in `tests/test_rebel_loss.py`

## Design rationale, experiments, and write-up

Full design doc covering motivation, implementation choices, experiment setup, and DPO/IPO/REBEL comparison:

- **Design doc:** https://github.com/LeoPhilly/supplement-rebel/blob/main/design_doc.md
- **Preprocessing script:** https://github.com/LeoPhilly/supplement-rebel/blob/main/preprocessing.py
- **Wandb charts:** https://github.com/LeoPhilly/supplement-rebel/blob/main/Wandb_charts.pdf

## Sanity-check experiment

Three runs at identical config (Qwen2.5-1.5B-Instruct + LoRA, 4096 train samples, 1631 held-out eval, seed 42), differing only in loss-specific flags:

| Loss | eval acc | final train loss\* | final grad norm |
|---|---|---|---|
| DPO | 0.730 | 0.516 | 7.81 |
| IPO | 0.686 | 17.34 | 1102.46 |
| REBEL (η=3.333) | 0.719 | 11.31 | 1868.37 |

\* Loss magnitudes are not directly comparable across objectives.

This is intended as implementation sanity-check rather than benchmark — see the design doc for full details.

## Testing

- `python -m pytest tests/test_rebel_loss.py` — 6 unit tests pass
- Smoke test on Qwen2.5-1.5B-Instruct + UltraFeedback (256 samples) trains end-to-end
- 4096-sample sanity comparison runs vs DPO and IPO baselines (see design doc)


LeoPhilly:
Addressed gemini bot feedback comments!

PR #1252, opened 2026-06-18
fix: use action_mask.sum() for response_length to exclude tool-call tokens in multi-turn training

### LINKED ISSUE #1243: overlong_penalty should exclude tool response tokens from length calculation (use action_mask/action_ranges)

KamayaniR:
Fixes #1243

## Problem

In `_process_response_into_experience`, `response_length` was computed as the **span** from the first to the last action token:

\`\`\`python
ones_indices = torch.where(action_mask)[0]
response_length = (ones_indices[-1] - ones_indices[0] + 1).item() if len(ones_indices) else 0
\`\`\`

For single-turn training this is fine — the action span is contiguous, so span == sum.

For **multi-turn agentic training** (e.g. `MultiTurnAgentExecutor`), `action_mask` has gaps between turns where tool-call tokens sit. The span overcounts those gaps:

\`\`\`
action_mask: [0 0 0 | 1 1 1 1 | 0 0 0 | 1 1 1 1 1]
                       turn1      tool    turn2
span = 14-3+1 = 12   ← includes 3 tool tokens
sum  = 4+5    = 9    ← only model-generated tokens  ✓
\`\`\`

This causes `apply_overlong_penalty` to measure a longer-than-actual response and trigger false penalties on responses that are not actually overlong.

## Fix

Replace the span with `action_mask.sum()`, which matches the field's documented semantics (`"number of generated tokens"`) and is correct for both single-turn and multi-turn:

\`\`\`python
response_length = action_mask.sum().item()
\`\`\`

## Tests

Added `tests/test_response_length.py` with four cases:
- single-turn (both approaches agree — no regression)
- multi-turn with two turns separated by tool tokens (sum gives 9, span gives 12)
- single-token turns with a gap (sum gives 2, span gives 3)
- empty action mask (both give 0)

### DIFF
--- a/openrlhf/trainer/ppo_utils/samples_generator.py
+++ b/openrlhf/trainer/ppo_utils/samples_generator.py
@@ -273,8 +273,8 @@ class SamplesGenerator:
             rollout_log_probs = None
 
         # Collect simple stats about lengths and clipping.
-        ones_indices = torch.where(action_mask)[0]
-        response_length = (ones_indices[-1] - ones_indices[0] + 1).item() if len(ones_indices) else 0
+        # Use sum() not span, so tool-call tokens between turns are excluded.
+        response_length = action_mask.sum().item()
         total_length = attention_mask.float().sum()
         is_clipped = total_length >= truncate_length
 
--- a/openrlhf/utils/deepspeed/deepspeed_utils.py
+++ b/openrlhf/utils/deepspeed/deepspeed_utils.py
@@ -141,7 +141,12 @@ def get_optimizer_grouped_parameters(
             "weight_decay": 0.0,
         },
     ]
-    return optimizer_grouped_parameters
+    # Filter empty groups: DeepSpeed removes empty param groups from the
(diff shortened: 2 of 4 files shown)


KamayaniR:
Reopening from a dedicated branch to keep this PR isolated from unrelated commits.

PR #1253, opened 2026-06-18
fix: filter empty param groups to prevent torch 2.10 LoRA LR scheduler crash

### LINKED ISSUE #1225: Torch 2.10's `LRScheduler._update_lr` breaks LORA
DeepSpeed's `_configure_optimizer` filters empty param groups out of the underlying optimizer in place (engine.py:1576), but the LR scheduler is built before `strategy.prepare()` runs and captures base_lrs from the  original group count. The length mismatch trips torch 2.10's
    [`zip(..., strict=True)`](https://github.com/pytorch/pytorch/blob/v2.10.0/torch/optim/lr_scheduler.py#L296) in `LRScheduler._update_lr` with "zip() argument 2 is longer than argument 1" at the first scheduler step.

Affects LoRA SFT: adapter param names (`lora_A.default.weight`,  `lora_B.default.weight`), when configured with weight_decay, match none of the no-decay patterns, so the no-decay group is always empty. Filtering at construction keeps the scheduler and optimizer in sync from the start.

The fix is a single line in `openrlhf/utils/deepspeed/deepspeed_utils.py`:

```
def get_optimizer_grouped_parameters(
    model,
    weight_decay,
    no_decay_name_list=["bias", "layer_norm.weight", "layernorm.weight", "norm.weight", "ln_f.weight"],
):
    optimizer_grouped_parameters = [
        {
            "params": [
                p
                for n, p in model.named_parameters()
                if (not any(nd in n for nd in no_decay_name_list) and p.requires_grad)
            ],
            "weight_decay": weight_decay,
        },
        {
            "params": [
                p
                for n, p in model.named_parameters()
                if (any(nd in n for nd in no_decay_name_list) and p.requires_grad)
            ],
            "weight_decay": 0.0,
        },
    ]

    # Drop empty groups: DeepSpeed's _configure_optimizer mutates
    # basic_optimizer.param_groups in place to remove groups with no params, but
    # the LR scheduler is constructed before that and captures base_lrs from the
    # original group count. The resulting length mismatch trips torch 2.10's
    # zip(..., strict=True) in LRScheduler._update_lr. This hits LoRA runs where
    # adapter param names (lora_A/lora_B) match none of the no-decay patterns,
    # leaving the second group empty.
    return [g for g in optimizer_grouped_parameters if g["params"]]
```

KamayaniR:
Fixes #1225

## Problem

`get_optimizer_grouped_parameters` always returns two groups (with-decay and no-decay). When running LoRA SFT, adapter params (`lora_A`, `lora_B`) never match any pattern in `no_decay_name_list`, so the zero-weight-decay group is always empty.

DeepSpeed's `_configure_optimizer` then strips empty groups from the underlying optimizer **in-place** — but the LR scheduler was already constructed and captured `base_lrs` from the original two-group count.

Torch 2.10 added `strict=True` to the `zip` inside `LRScheduler._update_lr`:
```
zip() argument 2 is longer than argument 1
```
This crashes on the very first scheduler step for any LoRA run on torch 2.10+.

## Fix

One line added at the end of `get_optimizer_grouped_parameters`:

```python
return [g for g in optimizer_grouped_parameters if g["params"]]
```

This keeps the scheduler and optimizer in sync from the start. Has no effect when both groups are non-empty (standard full-parameter training).

## Tests

Added `tests/test_optimizer_grouped_parameters.py`:
- Full model (linear with bias): both groups returned, no change
- LoRA-style model (only adapter params, no bias/norm): empty no-decay group filtered, one group returned
- Invariant: no returned group is ever empty

### DIFF
--- a/openrlhf/trainer/ppo_utils/samples_generator.py
+++ b/openrlhf/trainer/ppo_utils/samples_generator.py
@@ -273,8 +273,8 @@ class SamplesGenerator:
             rollout_log_probs = None
 
         # Collect simple stats about lengths and clipping.
-        ones_indices = torch.where(action_mask)[0]
-        response_length = (ones_indices[-1] - ones_indices[0] + 1).item() if len(ones_indices) else 0
+        # Use sum() not span, so tool-call tokens between turns are excluded.
+        response_length = action_mask.sum().item()
         total_length = attention_mask.float().sum()
         is_clipped = total_length >= truncate_length
 
--- a/openrlhf/utils/deepspeed/deepspeed_utils.py
+++ b/openrlhf/utils/deepspeed/deepspeed_utils.py
@@ -141,7 +141,12 @@ def get_optimizer_grouped_parameters(
             "weight_decay": 0.0,
         },
     ]
-    return optimizer_grouped_parameters
+    # Filter empty groups: DeepSpeed removes empty param groups from the
+    # optimizer in-place, but the LR scheduler captures base_lrs before that.
+    # Torch 2.10+ uses strict=True in LRScheduler._update_lr, so the length
+    # mismatch crashes on the first step. LoRA hits this because adapter params
+    # never match no_decay_name_list, leaving the zero-weight-decay group empty.
+    return [g for g in optimizer_grouped_parameters if g["params"]]
 
 
 def _z3_params_to_fetch(param_list):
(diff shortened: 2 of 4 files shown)


KamayaniR:
Closing in favour of a clean branch — the previous branch accidentally included commits from a separate PR (#1252).

PR #1268, opened 2026-07-21
test: run loss aggregation checks on accelerators

Madhustat:
Relates to https://github.com/OpenRLHF/OpenRLHF/issues/1303
## Summary

Extend `tests/test_loss_aggregation.py` to run the existing loss-aggregation test suite on CPU and, when available, the accelerator detected by PyTorch.

The tests currently create tensors without an explicit device, so they execute only on CPU. While this validates the mathematical logic, it does not verify numerical behavior, operator support, or tensor placement on the accelerator used during actual training.

This change adds accelerator coverage through PyTorch's generic accelerator API while retaining CPU coverage and avoiding separate CUDA-, XPU-, or vendor-specific copies of the tests.

## Changes

- Parameterize all eight existing loss-aggregation tests over:
  - CPU, which is always included.
  - The available accelerator reported by `torch.accelerator`, when present.
- Create input, mask, expected-result, log-probability, and advantage tensors directly on the selected device.
- Synchronize accelerator operations before validating results.
- Verify that relevant output tensors remain on the selected device.
- Preserve the original mathematical assertions and behavioral coverage of all eight tests.
- Keep the tests backend-neutral, without separate CUDA- or XPU-specific implementations.

## Motivation

During OpenRLHF training, loss, log-probability, advantage, and mask tensors normally reside on the training accelerator. CPU-only tests validate the formulas but do not exercise the same device path used in real training.

An implementation can behave correctly on CPU while still failing on an accelerator because of:

- Unsupported device operations.
- Accidental creation of CPU tensors.
- Device-mismatch errors.
- Unexpected movement of outputs back to CPU.
- Accelerator-specific numerical or reduction behavior.

Running the same tests on CPU and the available accelerator validates both mathematical correctness and tensor placement without duplicating the test suite for individual hardware vendors.

CPU-only environments remain unaffected because CPU is always included and accelerator parameterizations are added only when PyTorch reports an available accelerator.

## Test Coverage

The eight existing test functions continue to validate:

- Token-level mean aggregation across data-parallel partitions.
- Sequence-level mean aggregation across data-parallel partitions.
- Exclusion of fully masked sequences from batch-size calculation.
- Policy-loss token-mean aggregation.
- PPO and vLLM KL metric calculations.
- Global token normalization across gradient-accumulation windows.
- The difference between global and per-microbatch normalization for uneven token counts.
- Unclamped policy KL metric behavior.

The change extends these existing checks to the selected accelerator and does not alter their intended mathematical behavior.

### Focused validation

```bash
python -m pytest tests/test_loss_aggregation.py -v
```

Results:

```text
Intel XPU:   16 passed - 8 CPU + 8 XPU
NVIDIA CUDA: 16 passed - 8 CPU + 8 CUDA
```

The focused validation confirms that all eight loss-aggregation tests:

- Preserve their original numerical behavior on CPU.
- Execute successfully on Intel XPU and NVIDIA CUDA.
- Keep relevant output tensors on the selected accelerator.
- Use the same backend-neutral test implementation across platforms.

### Full regression validation

```bash
python -m pytest tests -v
```

Results:

```text
Intel XPU:   25 passed, 0 failed
NVIDIA CUDA: 25 passed, 0 failed
```

This confirms that extending the loss-aggregation tests to accelerators does not regress the existing OpenRLHF test suite.

### Code-quality checks

```bash
pre-commit run --all-files
```

Result:

```text
All checks passed
```

## Scope

This PR modifies only:

```text
tests/test_loss_aggregation.py
```

It does not modify OpenRLHF training logic, loss-computation implementation, distributed initialization, Ray, DeepSpeed, or vLLM behavior.

### DIFF
--- a/tests/test_loss_aggregation.py
+++ b/tests/test_loss_aggregation.py
@@ -3,6 +3,7 @@ import sys
 import types
 from pathlib import Path
 
+import pytest
 import torch
 
 _TEST_PACKAGE = "_openrlhf_loss_test"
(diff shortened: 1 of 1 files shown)


Madhustat:
Hi @hijkzzz , @Excelius-Wang 

This PR adds accelerator-aware loss aggregation coverage used during our single and multi-XPU validation.
The change is test-only and helps protect distributed numerical correctness across backends. 
I’m happy to address any feedback needed to help move this PR forward. 

Regards,
Madhu

PR #1278, opened 2026-08-04
fix(reward): stabilize log-exp loss for large gaps

RerankerGuo:
## Background

`LogExpLoss` currently evaluates the pairwise objective as:

```python
torch.log(1 + torch.exp(reject_reward - chosen_reward))
```

This direct exponential overflows for large reward gaps. For example, a rejected-minus-chosen gap of `100` produces an infinite loss and a `NaN` gradient in float32, which can poison reward model training when `--model.loss_type` selects the log-exp path.

## Changes

- Evaluate the same objective with `torch.nn.functional.softplus`.
- Add a regression test that checks both the large-gap loss value and gradient finiteness.

`softplus(x)` is mathematically equivalent to `log(1 + exp(x))` while using a numerically stable implementation.

## Verification

Before:

```text
gap=100  loss=inf  grad=nan
gap=1000 loss=inf  grad=nan
```

After:

```text
gap=100  loss=100.0  grad=-1.0
gap=1000 loss=1000.0 grad=-1.0
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_loss_aggregation.py -q
# 9 passed

# 18 passed

.venv/bin/pre-commit run --files openrlhf/models/loss.py tests/test_loss_aggregation.py
# all hooks passed
```

The full local test run used a minimal `deepspeed` import stub because DeepSpeed is not available on macOS; no DeepSpeed behavior is exercised by these tests.

## Impact

- Breaking change: No
- Affected module: reward model pairwise log-exp loss
- Performance impact: None expected
- Scope: The default sigmoid loss and other training losses are unchanged

## Checklist

- [x] The change is focused on one numerical correctness issue.
- [x] Regression coverage includes forward and backward behavior.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/models/loss.py
+++ b/openrlhf/models/loss.py
@@ -294,7 +294,7 @@ class LogExpLoss(nn.Module):
     def forward(
         self, chosen_reward: torch.Tensor, reject_reward: torch.Tensor, margin: torch.Tensor = None
     ) -> torch.Tensor:
-        loss = torch.log(1 + torch.exp(reject_reward - chosen_reward)).mean()
+        loss = F.softplus(reject_reward - chosen_reward).mean()
         return loss
 
 
--- a/tests/test_loss_aggregation.py
+++ b/tests/test_loss_aggregation.py
@@ -38,6 +38,7 @@ def _load_loss_utils_module():
 
 _loss_module = _load_loss_module()
 _loss_utils_module = _load_loss_utils_module()
+LogExpLoss = _loss_module.LogExpLoss
 PolicyLoss = _loss_module.PolicyLoss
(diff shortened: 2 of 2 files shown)


RerankerGuo:
Revalidated this numerical-stability fix against the current `main` (`bc71bb1`):

- before: `loss=inf`, `grad=nan` for a reward gap of 1000
- after: `loss=1000.0`, `grad=-1.0`
- `tests/test_loss_aggregation.py`: 9 passed
- pre-commit on both changed files: all hooks passed
- branch remains current and mergeable

The local macOS test run used the minimal DeepSpeed import stub noted in the PR description; the regression itself exercises only the PyTorch loss path.

PR #1281, opened 2026-08-04
fix(logging): align W&B eval step metric

RerankerGuo:
## Background

`WandbLogger` defines PPO evaluation metrics with `eval/epoch` as their custom step:

```python
wandb.define_metric("eval/*", step_metric="eval/epoch", step_sync=True)
```

However, `log_eval` emits `eval/global_step` and never logs `eval/epoch`. The declared step metric therefore does not match the actual payload. The DPO, RM, and SFT trainers already use `eval/global_step`.

## Changes

- Define `eval/global_step` as the W&B evaluation step metric.
- Keep the existing `log_eval` payload unchanged.
- Add a fake-W&B regression test covering both metric definitions and the logged payload.

## Verification

Before:

```text
defined step metric: eval/epoch
logged step key:     eval/global_step
```

After:

```text
defined step metric: eval/global_step
logged step key:     eval/global_step
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_logging_utils.py -q
# 1 passed

# 18 passed

  openrlhf/utils/logging_utils.py \
  tests/test_logging_utils.py
# all hooks passed
```

## Impact

- Breaking change: No
- Affected module: PPO W&B evaluation logging
- Performance impact: None
- Scope: Training metrics and non-W&B loggers are unchanged

## Checklist

- [x] The change addresses one observability correctness issue.
- [x] Regression coverage verifies definitions and emitted keys.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/utils/logging_utils.py
+++ b/openrlhf/utils/logging_utils.py
@@ -78,8 +78,8 @@ class WandbLogger:
 
         wandb.define_metric("train/global_step")
         wandb.define_metric("train/*", step_metric="train/global_step", step_sync=True)
-        wandb.define_metric("eval/epoch")
-        wandb.define_metric("eval/*", step_metric="eval/epoch", step_sync=True)
+        wandb.define_metric("eval/global_step")
+        wandb.define_metric("eval/*", step_metric="eval/global_step", step_sync=True)
         self.handle = wandb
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #1283, opened 2026-08-04
fix(data): normalize local dataset extensions

RerankerGuo:
## Background

Local dataset format detection compares file extensions before normalizing their case. Uppercase or mixed-case paths such as `train.JSONL` therefore skip the local-file branch and are passed to Hugging Face as dataset repository names:

```text
load_dataset("/tmp/train.JSONL", data_dir=None)
```

File extensions are case-insensitive on common workflows, and the expected call is:

```text
load_dataset("json", data_files="/tmp/train.JSONL")
```

## Changes

- Normalize the local path extension before format dispatch.
- Preserve the existing JSONL-to-JSON loader mapping.
- Add regression coverage for uppercase JSON, JSONL, CSV, Parquet, and Arrow paths.

## Verification

```bash
.venv/bin/python -m pytest tests/test_dataset_utils.py -q
# 5 passed

# 22 passed

  openrlhf/datasets/utils.py \
  tests/test_dataset_utils.py
# all hooks passed
```

The full local test run used a minimal `deepspeed` import stub because DeepSpeed is not available on macOS. Dataset loading is mocked in the focused test to verify exact loader arguments.

## Impact

- Breaking change: No
- Affected module: local dataset loading
- Performance impact: None
- Scope: Lowercase paths and remote dataset names retain their existing behavior

## Checklist

- [x] The change addresses one compatibility issue.
- [x] Regression tests cover all supported local file formats.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/datasets/utils.py
+++ b/openrlhf/datasets/utils.py
@@ -40,7 +40,7 @@ def blending_datasets(
         dataset = dataset.split("@")[0].strip()
         dataset_basename = os.path.basename(dataset)
 
-        ext = os.path.splitext(dataset)[-1]
+        ext = os.path.splitext(dataset)[-1].lower()
         # local python script
         if ext == ".py" or (
             os.path.isdir(dataset) and os.path.exists(os.path.join(dataset, f"{dataset_basename}.py"))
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #1284, opened 2026-08-04
fix(vlm): release image sources after loading

RerankerGuo:
## Background

`load_images` returns PIL objects created with `Image.open` from paths and in-memory streams. PIL loads lazily, so those returned objects retain their backing file handles until callers explicitly close them.

Loading the same local image 20 times currently returns 20 images with 20 open `image.fp` handles. Long-running VLM rollouts can therefore accumulate descriptors.

## Changes

- Materialize a pixel copy for images opened by `load_images`.
- Close path and in-memory sources before returning.
- Apply the same lifecycle to URL, bytes, and base64 image inputs.
- Preserve ownership of caller-provided `PIL.Image` objects.
- Add regression coverage for handle closure and caller-owned images.

## Verification

Before:

```text
images=20 open_file_handles=20
```

After:

```text
images=20 open_file_handles=0
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_vlm_image_loading.py -q
# 2 passed

# 19 passed

  openrlhf/utils/vlm_utils.py \
  tests/test_vlm_image_loading.py
# all hooks passed
```

## Impact

- Breaking change: No
- Affected module: VLM image loading
- Performance impact: Eagerly materializes images that downstream processors already consume
- Scope: Caller-provided PIL image ownership is unchanged

## Checklist

- [x] The change addresses one resource lifecycle issue.
- [x] Regression tests cover handle closure and ownership.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/utils/vlm_utils.py
+++ b/openrlhf/utils/vlm_utils.py
@@ -27,6 +27,12 @@ def _is_base64_image(s: str) -> bool:
     return False
 
 
+def _open_image_copy(source) -> Image.Image:
+    """Open and materialize an image without retaining the source resource."""
+    with Image.open(source) as image:
+        return image.copy()
+
+
 def load_images(image_refs: Union[str, List[str], Image.Image, List[Any]]) -> List[Image.Image]:
     """Load PIL images from paths, URLs, base64 strings, raw bytes, or PIL objects.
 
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #1285, opened 2026-08-04
fix(reward): extract answers from fbox commands

RerankerGuo:
## Background

The math answer extractor searches for both `\boxed{}` and `\fbox{}` commands, but `remove_boxed` only accepts the `\boxed{` prefix. As a result:

```text
extract_boxed_answer(r"\boxed{42}") -> "42"
extract_boxed_answer(r"\fbox{42}")  -> None
```

This silently drops otherwise valid math answers before reward grading.

## Changes

- Accept both `\boxed{}` and `\fbox{}` prefixes during answer removal.
- Preserve malformed-input behavior (`None`).
- Update the public helper docstring.
- Add coverage for scalar, nested fraction, and unclosed commands.

## Verification

After the change:

```text
\boxed{42}             -> 42
\fbox{42}              -> 42
\fbox{\frac{1}{2}}     -> \frac{1}{2}
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_boxed_answer_extraction.py -q
# 4 passed

# 21 passed

  openrlhf/utils/math_utils.py \
  tests/test_boxed_answer_extraction.py
# all hooks passed
```

## Impact

- Breaking change: No
- Affected module: math reward answer extraction
- Performance impact: None
- Scope: Existing `\boxed{}` behavior is unchanged

## Checklist

- [x] The change addresses one reward correctness issue.
- [x] Regression tests cover valid and malformed commands.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/utils/math_utils.py
+++ b/openrlhf/utils/math_utils.py
@@ -353,17 +353,14 @@ def last_boxed_only_string(string: str) -> str | None:
 
 
 def remove_boxed(value: str) -> str | None:
-    left = "\\boxed{"
-    try:
-        assert value[: len(left)] == left
-        assert value[-1] == "}"
-        return value[len(left) : -1]
-    except Exception:
-        return None
+    for left in ("\\boxed{", "\\fbox{"):
+        if value.startswith(left) and value.endswith("}"):
+            return value[len(left) : -1]
+    return None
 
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #1286, opened 2026-08-04
fix(checkpoint): normalize legacy PPO state

RerankerGuo:
## Background

PPO resume assumes every loaded checkpoint contains all current controller fields:

```python
checkpoint_states["total_consumed_prompts"]
checkpoint_states["data_loader_state_dict"]
```

Legacy checkpoints and model-only checkpoints may only contain `episode` and `global_step`. Model loading then succeeds, but controller startup fails with `KeyError`.

The async path also passes an empty dataloader state to `load_state_dict` whenever `global_step > 0`.

## Changes

- Normalize loaded checkpoint state with defaults for current controller fields.
- Preserve all fields that are present in the checkpoint.
- Warn when an actual loaded checkpoint is missing controller fields.
- Avoid calling async dataloader restoration when no state exists.
- Add regression coverage for legacy-state normalization and loaded-state integration.

Missing dataloader state cannot reconstruct a historical cursor; the warning makes that limitation explicit while allowing model/controller recovery to proceed from the loader's default position.

## Verification

Before:

```text
{"episode": 1, "global_step": 3}
-> KeyError: "total_consumed_prompts"
-> KeyError: "data_loader_state_dict"
```

After:

```text
{
  "episode": 1,
  "global_step": 3,
  "total_consumed_prompts": 0,
  "data_loader_state_dict": {}
}
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_ppo_checkpoint_state.py -q
# 2 passed

# 19 passed

  openrlhf/trainer/ppo_trainer.py \
  openrlhf/trainer/ppo_trainer_async.py \
  tests/test_ppo_checkpoint_state.py
# all hooks passed
```

## Impact

- Breaking change: No
- Affected module: sync and async PPO checkpoint recovery
- Performance impact: None
- Scope: Complete current-format checkpoints retain identical state

## Checklist

- [x] The change addresses one checkpoint compatibility issue.
- [x] Regression tests cover legacy and integration paths.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/trainer/ppo_trainer.py
+++ b/openrlhf/trainer/ppo_trainer.py
@@ -23,6 +23,21 @@ from openrlhf.utils.utils import get_tokenizer
 logger = init_logger(__name__)
 
 
+def normalize_checkpoint_states(checkpoint_states=None):
+    """Fill controller fields missing from legacy or model-only checkpoints."""
+    defaults = {
+        "episode": 0,
+        "global_step": 0,
+        "total_consumed_prompts": 0,
+        "data_loader_state_dict": {},
+    }
+    states = dict(checkpoint_states or {})
+    missing = [key for key in defaults if key not in states]
+    if checkpoint_states is not None and missing:
+        logger.warning(f"Checkpoint is missing controller state {missing}; using defaults.")
+    return {**defaults, **states}
+
+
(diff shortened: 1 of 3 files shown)


RerankerGuo:

PR #1287, opened 2026-08-04
fix(batch): align dynamic micro-batches upward

RerankerGuo:
## Background

Dynamic rollout batching first computes the minimum number of micro-batches needed to respect the token capacity, then aligns that number to the effective actor count:

```python
minimum_batch_num = minimum_batch_num // effective_actor_num * effective_actor_num
```

This rounds down. If the capacity calculation requires 5 micro-batches and there are 4 effective actors, the result becomes 4, violating the already-computed minimum and potentially exceeding the per-GPU token limit.

## Changes

- Round the minimum micro-batch count up to an effective-actor multiple.
- Keep at least one non-empty batch per effective actor.
- Reject configurations where the aligned count exceeds the sample count and would require empty partitions.
- Add regression coverage for upward alignment, actor minimum, and infeasible alignment.

## Verification

Before:

```text
minimum=5, actors=4 -> aligned=4
```

After:

```text
minimum=5, actors=4, samples=8 -> aligned=8
minimum=3, actors=4, samples=4 -> aligned=4
minimum=5, actors=4, samples=5 -> explicit ValueError
```

Commands:

```bash
.venv/bin/python -m pytest tests/test_dynamic_batch_alignment.py -q
# 3 passed

# 20 passed

  openrlhf/utils/seqlen_balancing.py \
  openrlhf/trainer/ppo_utils/experience_maker.py \
  tests/test_dynamic_batch_alignment.py
# all hooks passed
```

## Impact

- Breaking change: No for feasible configurations
- Affected module: dynamic rollout micro-batch partitioning
- Performance impact: May use more micro-batches when required to preserve the configured token limit
- Scope: Already aligned counts are unchanged

## Checklist

- [x] The change addresses one dynamic-batch correctness issue.
- [x] Regression tests cover feasible and infeasible boundaries.
- [x] Pre-commit hooks pass.

### DIFF
--- a/openrlhf/trainer/ppo_utils/experience_maker.py
+++ b/openrlhf/trainer/ppo_utils/experience_maker.py
@@ -11,7 +11,11 @@ from openrlhf.trainer.ppo_utils.experience import Experience
 from openrlhf.trainer.ppo_utils.length_penalty import apply_length_penalties
 from openrlhf.trainer.ray.launcher import RayActorGroup
 from openrlhf.utils.logging_utils import init_logger
-from openrlhf.utils.seqlen_balancing import get_minimum_num_micro_batch_size, get_seqlen_balanced_partitions
+from openrlhf.utils.seqlen_balancing import (
+    align_num_micro_batches,
+    get_minimum_num_micro_batch_size,
+    get_seqlen_balanced_partitions,
+)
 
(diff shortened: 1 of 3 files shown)


RerankerGuo:

PR #1293, opened 2026-08-05
fix(gspo): bound the importance ratio exponent so the gradient stays finite

adityasingh2400:
## What

`PolicyLoss.forward` computes a `clamp(-20, 20)` guard in the `gspo` branch and then never reads it. `openrlhf/models/loss.py`, before this PR:

```python
elif self.policy_loss_type == "gspo":
    if self.enable_vllm_is_correction:
        log_ratio = log_probs - rollout_log_probs
    else:
        log_ratio = raw_policy_log_ratio
    policy_log_ratio = raw_policy_log_ratio.clamp(min=-20.0, max=20.0)   # assigned, never read
    ratio = (log_ratio * action_mask).sum(dim=-1) / action_mask.sum(dim=-1).clamp(min=1)
    ratio = ratio.exp().unsqueeze(-1) * action_mask                      # unclamped exponent
```

The `ppo` branch two lines above does consume its clamp (`ratio = policy_log_ratio.exp()`), so `gspo` is the only path that exponentiates an unbounded log ratio. `policy_log_ratio` is dead in this branch: the only other reader in the function is line 230, which uses `raw_policy_log_ratio`.

## Why it matters

With `--actor.policy_loss_type gspo`, a sequence whose mean log ratio is large makes `ratio` `inf`. The clipped objective then does `torch.min(surr1, surr2)` where one branch is `inf`, and the backward pass turns that into `0 * inf = nan`. Because every sequence in the batch is produced by the same actor weights, the nan is not confined to the offending sequence: it reaches every parameter gradient, so one bad sequence destroys the whole optimizer step. The forward loss value survives, so this shows up as a model that silently goes to nan rather than as an error at the loss.

Measured on `main`, `log_probs - old_log_probs = 100` on one of two sequences, everything else ordinary:

| policy_loss_type | loss finite | ratio | grad w.r.t. a shared weight |
|---|---|---|---|
| `ppo` | yes | `485165184.0` = exp(20), clamp applied | finite |
| `gspo` | yes | `inf` = exp(100) overflows fp32 | `nan` |

At a log ratio of 30 both paths are still finite, so the clamp is exactly what separates them.

This is reachable in normal use rather than only under a synthetic input. When `enable_vllm_is_correction` is set, line 173 makes `log_ratio = log_probs - rollout_log_probs`, a comparison between logprobs produced by two different engines, which is precisely the setting where large log ratios appear.

## Fix

Clamp the quantity that is actually exponentiated, which is the sequence mean, and drop the dead assignment:

```python
seq_log_ratio = (log_ratio * action_mask).sum(dim=-1) / action_mask.sum(dim=-1).clamp(min=1)
ratio = seq_log_ratio.clamp(min=-20.0, max=20.0).exp().unsqueeze(-1) * action_mask
```

Bounding the sequence mean rather than each token's log ratio is the minimal analogue of what the `ppo` branch guards: it keeps `exp()` representable without otherwise changing GSPO's importance ratio, so ratios that were already finite are bit-for-bit unchanged. There is a test asserting that.

## Second change, and why it is in the same PR

`openrlhf/models/utils.py` imported `deepspeed` at module scope for a single call site inside `set_z3_leaf_modules`, itself behind `if z3_leaf_classes:`. `tests/test_loss_aggregation.py` goes out of its way to load `models/utils.py` and `models/loss.py` by file path specifically to avoid the heavy dependencies, and that one import defeated the whole design: the file could not even be collected without DeepSpeed installed.

```
$ pytest tests/test_loss_aggregation.py -q
openrlhf/models/utils.py:4: in <module> import deepspeed
E   ModuleNotFoundError: No module named 'deepspeed'
collected 0 items / 1 error
```

The import is now function-local. This is what lets the regression tests below run on a plain CPU machine, which is why it is bundled here rather than split out. `set_z3_leaf_modules` behaviour is unchanged.

## Tests

Added to `tests/test_loss_aggregation.py`, reusing its existing file-path module loader. CPU only, no GPU, no DeepSpeed, no model download.

- `test_large_log_ratio_keeps_the_gradient_finite[ppo|gspo]`, both policy loss types must keep `loss` and the gradient finite on the same input
- `test_large_log_ratio_does_not_poison_shared_parameters[ppo|gspo]`, routes both sequences through one shared scalar parameter, which is what actually models an actor, and asserts that parameter's gradient stays finite
- `test_gspo_ratio_matches_the_unclamped_sequence_mean_in_normal_range`, the new bound must not perturb ratios that were already representable, checked against a direct reimplementation of the sequence-mean objective

```
$ pytest tests/test_loss_aggregation.py -q
13 passed in 3.80s
```

Fail-before and pass-after, reverting only `openrlhf/models/loss.py` and keeping the new tests:

```
$ git checkout origin/main -- openrlhf/models/loss.py
$ pytest tests/test_loss_aggregation.py -q
FAILED tests/test_loss_aggregation.py::test_large_log_ratio_keeps_the_gradient_finite[gspo]
FAILED tests/test_loss_aggregation.py::test_large_log_ratio_does_not_poison_shared_parameters[gspo]
2 failed, 11 passed

E   AssertionError: gspo gradient is not finite: tensor([[nan, nan], [0., 0.]])
E   AssertionError: gspo poisoned the shared gradient: tensor([nan])

$ git checkout HEAD -- openrlhf/models/loss.py
$ pytest tests/test_loss_aggregation.py -q
13 passed in 3.80s
```

The `ppo` parametrization of both tests passes before and after, which is what pins the `ppo`-vs-`gspo` asymmetry as the defect.

Formatting, at the versions pinned in `.pre-commit-config.yaml` (autoflake 2.3.3, isort 8.0.1, black 26.3.1 with click 8.0.2):

```
$ autoflake --remove-all-unused-imports --in-place openrlhf/models/loss.py openrlhf/models/utils.py tests/test_loss_aggregation.py
$ isort openrlhf/models/loss.py openrlhf/models/utils.py tests/test_loss_aggregation.py
$ black openrlhf/models/loss.py openrlhf/models/utils.py tests/test_loss_aggregation.py
All done! 3 files left unchanged.
```

## Not a duplicate

- `+is:pr+is:open+gspo`, `+policy_log_ratio`, `+gspo+clamp` return nothing
- open PRs touching `openrlhf/models/loss.py` are #1278, #1240, #1247 and #1184
- #1240 (`fix: ignore masked invalid values in policy loss reductions`) is the nearest neighbour and edits the same function, but it zeroes masked positions in the inputs before the ratio is computed. It does not touch lines 176 to 178, and it does not fix this: the overflow here comes from an unmasked token with a large log ratio, so the nan survives #1240 unchanged. The two are complementary and only textually adjacent.

## Notes

AI assistance was used to write this change. Every number above is from a run on this branch rather than an estimate: the `inf` ratio, the two nan gradients, the `ppo`-vs-`gspo` split at log ratio 20 versus 30, and the collection error. An earlier draft of this description claimed the nan reaches every sequence's `log_probs` gradient; that was wrong, only the offending row's leaf gradient is nan, which is why the shared-parameter test exists to show the damage that actually matters.

### DIFF
--- a/openrlhf/models/loss.py
+++ b/openrlhf/models/loss.py
@@ -173,9 +173,12 @@ class PolicyLoss(nn.Module):
                 log_ratio = log_probs - rollout_log_probs
             else:
                 log_ratio = raw_policy_log_ratio
-            policy_log_ratio = raw_policy_log_ratio.clamp(min=-20.0, max=20.0)
-            ratio = (log_ratio * action_mask).sum(dim=-1) / action_mask.sum(dim=-1).clamp(min=1)
-            ratio = ratio.exp().unsqueeze(-1) * action_mask
+            seq_log_ratio = (log_ratio * action_mask).sum(dim=-1) / action_mask.sum(dim=-1).clamp(min=1)
+            # Bound the exponent, as the ppo branch above does. Without this a sequence
+            # with a large log ratio makes ratio inf, and torch.min below then turns that
+            # into 0 * inf = nan in the backward pass. Since every sequence shares the
+            # actor weights, one such sequence is enough to nan the whole update.
+            ratio = seq_log_ratio.clamp(min=-20.0, max=20.0).exp().unsqueeze(-1) * action_mask
         else:
             raise ValueError(f"Invalid policy loss type: {self.policy_loss_type}")
 
--- a/openrlhf/models/utils.py
+++ b/openrlhf/models/utils.py
@@ -1,7 +1,6 @@
 import os
 from typing import Optional, Tuple, Union
 
-import deepspeed
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
@@ -56,6 +55,10 @@ def set_z3_leaf_modules(model: nn.Module, detect_hybrid: bool = True) -> None:
                 child_sigs[cls] = children
 
     if z3_leaf_classes:
+        # Imported here, not at module scope, so that the pure-torch helpers in this
+        # module stay importable without DeepSpeed installed.
+        import deepspeed
+
         deepspeed.utils.set_z3_leaf_modules(model, list(z3_leaf_classes))
         for cls in z3_leaf_classes:
             print(f"Setting zero3 leaf: {cls.__name__}")
--- a/tests/test_loss_aggregation.py
+++ b/tests/test_loss_aggregation.py
@@ -3,6 +3,7 @@ import sys
 import types
 from pathlib import Path
 
+import pytest
 import torch
 
 _TEST_PACKAGE = "_openrlhf_loss_test"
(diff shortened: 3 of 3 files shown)


Excelius-Wang (commented):
Clamping the sequence mean looks right here: clamping each token first would change GSPO's ratio even when the mean is in range.

I ran the CPU suite and added checks for both signs of advantage and the `rollout_log_probs` branch. With the old loss implementation, the extreme-ratio cases produce non-finite losses or gradients; they pass with this change. A sequence with log-ratios `[100, -100]` still gives ratio 1 and the expected gradient. I haven't tested this in a distributed training run.

adityasingh2400:
Thanks @Excelius-Wang. Rebased onto current main in `0accfcc` and added the checks you described in `4142894`, so the branch is mergeable again.

What conflicted: main renamed `enable_vllm_is_correction` to `is_correction_level` and appended new tests to `tests/test_loss_aggregation.py` after the fix was written. The `gspo` branch itself still exponentiated the unclamped sequence mean on main, so the fix is unchanged; only the surrounding lines moved. I dropped the earlier pre-commit.ci commit since it only reformatted `sft_dataset.py`, which main has since rewritten.

New tests, all in the same file:

- the two overflow tests are now parametrized over the sign of the overflowing sequence's advantage, so both the `surr1` and the clipped `surr2` side are covered for `ppo` and `gspo`
- `test_gspo_rollout_log_probs_branch_keeps_the_gradient_finite` sets `is_correction_level="token"` so `gspo` exponentiates `log_probs - rollout_log_probs`, with train and old logprobs equal and the rollout logprobs 100 off
- `test_gspo_cancelling_token_log_ratios_still_give_ratio_one` pins your `[100, -100]` case: loss is exactly -1, `clip_ratio` is 0, and the gradient is `[-0.5, -0.5]`. It also checks that `[100, -60]` gives `exp(20)`, since that is what separates bounding the mean from bounding each token.

Run on CPU with torch 2.14.0, no DeepSpeed:

```
$ pytest tests/test_loss_aggregation.py -q
25 passed in 1.57s
```

With only `openrlhf/models/loss.py` reverted to main:

```
FAILED test_large_log_ratio_keeps_the_gradient_finite[gspo-1.0]
FAILED test_large_log_ratio_keeps_the_gradient_finite[gspo--1.0]
FAILED test_large_log_ratio_does_not_poison_shared_parameters[gspo-1.0]
FAILED test_large_log_ratio_does_not_poison_shared_parameters[gspo--1.0]
FAILED test_gspo_rollout_log_probs_branch_keeps_the_gradient_finite
5 failed, 20 passed
```

The cancelling case passes on both, which is expected since it is a no-change guard rather than a regression test. autoflake, isort and black at the pinned versions leave the three files unchanged.

Not run: any distributed or GPU training, same as before.

PR #1302, opened 2026-08-12
make flash_attn utility imports optional

Madhustat:
Relates to #1303
## Summary

Make the external `flash_attn` utility imports in `openrlhf/models/ring_attn_utils.py` optional.

OpenRLHF currently imports padding and distributed utilities from `flash_attn` at module load time. The package requires platform-specific compiled CUDA or ROCm components and is unavailable in  other environment for example Intel XPU.

As a result, importing `openrlhf.models` fails before the ring-attention or packed-sample paths are exercised.

When `flash_attn` imports successfully, OpenRLHF continues using the existing FlashAttention implementations. When the `flash_attn` package is absent, the module uses backend-neutral equivalents based on existing OpenRLHF dependencies:

- `rearrange` directly from `einops`
- `pad_input` and `unpad_input` from Transformers
- A local axis-0 indexing helper
- An autograd-aware `torch.distributed` all-gather implementation

The fallback activates only when the top-level `flash_attn` package is absent. Import errors from a present but incomplete or incompatible FlashAttention installation are re-raised rather than hidden.

## Motivation

The utilities imported from `flash_attn` are required by OpenRLHF's packed tensor and ring-attention processing. Importing them unconditionally makes the complete models package depend on a compiled FlashAttention installation.

This prevents affected OpenRLHF modules from loading on platforms where `flash_attn` is unavailable, even when the current workflow does not execute the ring-attention or packed-sample paths.

This change makes dependency loading conditional while preserving the existing FlashAttention implementation whenever the package is available.

## Changes
when the package is available it imports `flash_attn` package. when the package is unavailable uses packages as listed down below.
### `rearrange`

Import `rearrange` directly from the existing `einops` dependency:

```python
from einops import rearrange
```

`flash_attn` re-exports this symbol, but the implementation belongs to `einops`. Making the import unconditional avoids coupling `rearrange` to FlashAttention availability.

### `pad_input` and `unpad_input`

When `flash_attn` is unavailable, use the corresponding Transformers
helpers:

```python
from transformers.modeling_flash_attention_utils import (
    _pad_input as pad_input,
    _unpad_input as unpad_input,
)
```

These helpers preserve the padding and unpadding contracts required by the existing OpenRLHF call sites.

The Transformers APIs are underscore-prefixed and therefore private.
Focused regression tests cover their expected behavior so that incompatible
upstream changes fail during testing.

### `index_first_axis`

The Transformers helper expects an unflattened `(batch, sequence, ...)` input and flattens the first two dimensions before indexing.

OpenRLHF passes an already-flattened tensor with shape:

```text
(batch * sequence, 1)
```

The FlashAttention contract selects rows along axis 0 while preserving all trailing dimensions. The fallback therefore uses:

```python
def index_first_axis(tensor, indices):
    return torch.index_select(tensor, 0, indices)
```

This preserves the expected result shape:

```text
Input:  (batch * sequence, 1)
Output: (selected_tokens, 1)
```

Preserving the trailing dimension is required by the subsequent transpose in the packed-sample path.

`torch.index_select` also preserves gradient propagation to the selected input rows.

### `all_gather`

Transformers does not provide an equivalent to `flash_attn.utils.distributed.all_gather`.

The fallback therefore provides an autograd-aware implementation using `torch.distributed`:

```text
Forward:
all_gather_into_tensor

Backward:
reduce_scatter_tensor
```

This preserves the distributed-autograd relationship expected by the existing OpenRLHF code.

## Exception Handling

The fallback activates only when the top-level  `flash_attn` package  is genuinely absent:

```python
except ModuleNotFoundError as exc:
    if exc.name != "flash_attn":
        raise
```

This distinction is intentional.

A missing top-level `flash_attn` package activates the fallback:

```text
ModuleNotFoundError: No module named 'flash_attn'
```

An installed but incomplete or incompatible package may instead report a missing internal module, such as:

```text
flash_attn_2_cuda
flash_attn.bert_padding
```

Those errors are re-raised so that a broken FlashAttention installation is not silently hidden.

## Files Changed

This PR modifies two files:

```text
openrlhf/models/ring_attn_utils.py
tests/test_ring_attn_utils.py
```

### `openrlhf/models/ring_attn_utils.py`

- Adds optional FlashAttention import handling
- Adds backend-neutral fallback implementations
- Imports `rearrange` directly from `einops`
- Preserves errors from incomplete FlashAttention installations

The file also contains one related device-portability substitution:

```text
torch.cuda.current_device()
    -> torch.accelerator.current_device_index()
```

This substitution removes a remaining CUDA-specific device lookup from the affected ring-attention path on non-CUDA accelerators.

### `tests/test_ring_attn_utils.py`

Adds focused regression coverage for the optional-import and fallback paths.

### Import without FlashAttention

With `flash_attn` unavailable, importing OpenRLHF models succeeds:

```bash
python -c \
  "import openrlhf.models; print('import OK without flash_attn')"
```

Expected output:

```text
import OK without flash_attn
```

### Focused Tests

Run:

```bash
python -m pytest tests/test_ring_attn_utils.py -v
```

The focused tests verify:

- `index_first_axis` preserves trailing dimensions
- The exact `(batch * sequence, 1)` regression shape is preserved
- Gradients flow to the selected input rows
- Packed-sample padding and unpadding preserve expected values and shapes
- Fallback symbols are selected when `flash_attn` is absent
- A missing top-level `flash_attn` package activates the fallback
- Missing internal FlashAttention modules are re-raised
- The distributed all-gather fallback preserves forward results
- The distributed all-gather fallback preserves backward gradients

The fallback-specific tests run when `flash_attn` is unavailable.

When `flash_attn` is installed, OpenRLHF selects the existing FlashAttention implementations instead, so the fallback-specific tests are skipped. The FlashAttention-installed path is unchanged structurally but is not exercised by these fallback-specific tests.

### Code-Quality Checks

Run:

```bash
pre-commit run --all-files
```

Result:

```text
All checks passed
```

### FlashAttention Installed

When the `flash_attn` imports succeed,  for example in the case of CUDA or ROCm,  OpenRLHF continues using the existing FlashAttention implementations.

The fallback path is not entered.

### FlashAttention Unavailable

When the top-level  `flash_attn` package is absent, `openrlhf.models` can be imported and the validated utility paths use the backend-neutral fallback implementations.

### Incomplete FlashAttention Installation

Errors from missing internal modules or compiled extensions are re-raised.

The fallback does not hide an installed but incomplete or incompatible FlashAttention package.

### Intel XPU

The module can be imported without a compiled FlashAttention installation.

The validated fallback helpers use existing PyTorch, Transformers, `einops`, and `torch.distributed` functionality.

## Scope

This PR changes:

- FlashAttention utility import handling
- Fallback padding and indexing helpers
- The distributed all-gather fallback
- One related device-portability lookup in the same ring-attention path
- Focused regression tests

It does not change:

- Attention kernels
- Model architecture
- PPO or GRPO algorithms
- Loss calculations
- Ray scheduling
- DeepSpeed configuration
- vLLM behavior
- Distributed-backend selection

No new dependency is introduced. `einops` and Transformers are already OpenRLHF dependencies.

## Test Checklist

- [x] `openrlhf.models` imports without `flash_attn`
- [x] Missing top-level `flash_attn` activates the fallback
- [x] Missing internal FlashAttention modules are re-raised
- [x] `index_first_axis` preserves trailing dimensions
- [x] `index_first_axis` gradient propagation is validated
- [x] Packed-sample utility round-trip is validated
- [x] Fallback-symbol selection is validated
- [x] Distributed all-gather forward behavior is validated
- [x] Distributed all-gather backward behavior is validated
- [x] `pre-commit run --all-files` passes

### DIFF
--- a/openrlhf/models/ring_attn_utils.py
+++ b/openrlhf/models/ring_attn_utils.py
@@ -1,7 +1,88 @@
 import torch
 import torch.distributed as dist
-from flash_attn.bert_padding import index_first_axis, pad_input, rearrange, unpad_input
-from flash_attn.utils.distributed import all_gather
+
+# rearrange lives in einops (an existing OpenRLHF dependency); flash_attn only re-exported it.
+from einops import rearrange
+
+# Prefer the external flash_attn implementations when the package is installed and usable
+# (the established NVIDIA/CUDA path, kept unchanged). Fall back to backend-neutral equivalents
+# ONLY when the top-level flash_attn package is absent - e.g. Intel XPU environments where the
+# flash_attn package is not available. We match exc.name exactly against "flash_attn": a
+# flash_attn that is installed-but-broken raises ModuleNotFoundError for one of its OWN
+# submodules (e.g. "flash_attn_2_cuda" when the compiled extension is missing, or
+# "flash_attn.bert_padding" on a version/layout mismatch), and those must surface as real
+# errors rather than silently activating the fallback.
+try:
+    from flash_attn.bert_padding import index_first_axis, pad_input, unpad_input
+    from flash_attn.utils.distributed import all_gather
+except ModuleNotFoundError as exc:
+    if exc.name != "flash_attn":
+        raise
+
+    # transformers ships PyTorch implementations of the padding helpers. They are private
+    # (underscore-prefixed) APIs, so tests/test_ring_attn_utils.py guards their expected
+    # signatures/behavior against transformers upgrades.
+    #
+    # _pad_input / _unpad_input match flash_attn's pad_input / unpad_input contracts and are
+    # aliased directly. _index_first_axis is NOT aliased: transformers' version assumes a
+    # (batch, seqlen, ...) input and flattens the first two dims before indexing, whereas
+    # flash_attn's index_first_axis selects rows along axis 0 and preserves all trailing dims.
+    # OpenRLHF calls index_first_axis on an already-flattened (batch*seqlen, 1) tensor and then
+    # transposes, which needs the trailing dim to survive - so we provide a flash_attn-compatible
+    # wrapper below instead. (Discovered via a real GRPO --packing_samples run, not unit tests.)
+    from transformers.modeling_flash_attention_utils import _pad_input as pad_input
+    from transformers.modeling_flash_attention_utils import _unpad_input as unpad_input
+
+    def index_first_axis(tensor, indices):
+        """Select rows along axis 0 while preserving all trailing dimensions.
+
+        Preserves the contract of flash_attn.bert_padding.index_first_axis:
+        (total, ...) -> (len(indices), ...). Pure torch, no CUDA-specific code; the
+        gather is autograd-aware so gradients flow back to the selected rows.
+        """
+        return torch.index_select(tensor, 0, indices)
+
+    class _AllGatherFunc(torch.autograd.Function):
+        """Autograd-aware all-gather fallback for flash_attn.utils.distributed.all_gather.
+
+        transformers ships no replacement for this one, so it is vendored here. Pure
+        torch.distributed, no CUDA-specific code. Forward all-gathers along dim 0 across ranks;
+        backward reduce-scatters so each rank receives the gradient for its original shard -
+        the same autograd relationship as flash_attn's version.
+        """
+
+        @staticmethod
+        def forward(ctx, input_, process_group):
+            ctx.process_group = process_group
+            if not dist.is_initialized():
+                raise RuntimeError("torch.distributed must be initialized before all_gather")
+            world_size = dist.get_world_size(process_group)
+            input_ = input_.contiguous()
+            output = torch.empty(
+                world_size * input_.shape[0], *input_.shape[1:], dtype=input_.dtype, device=input_.device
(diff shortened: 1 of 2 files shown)


Excelius-Wang (commented):
I checked the fallback utility module in isolation with the pinned Transformers 5.7.0. The packing round-trip tests pass for FP32 and BF16, including gradients. An additional two-rank CPU/Gloo check produced the expected summed gradients on both ranks.

I also injected missing-extension and missing-submodule errors into the actual imports; both are re-raised rather than activating the fallback. That distinction looks correct, and `index_select` preserves the trailing dimension needed by the packing call sites.

I didn't find a blocking issue in those utilities. I bypassed the models package initializer for these checks, so this doesn't verify the full package import, XPU/CUDA behavior, or a training run.

Madhustat:
Hi @hijkzzz, @Excelius-Wang

This PR removes the mandatory import dependency on CUDA-specific attention packages while preserving the existing behavior when they are installed.
We have validated the upstream workflows on single and multi XPU with this change. 
Please let me know if any changes would help move this PR through review and merge.

Regards,
Madhu

PR #1309, opened 2026-08-14
Fix reverse-KL gradient on the KL-as-loss path (form + off-policy importance weight)

anas353:
## Problem

Research on the role of the KL term in post-training particularly diversity collapse and reward hacking often turns on which divergence is actually being used as regularization: the reverse KL D_KL(pi_theta || pi_ref) or the forward KL D_KL(pi_ref || pi_theta). The current implementation differentiates k1, k2 or k3 directly, treating each as a gradient estimator for the reverse KL the config requests. None of them is unbiased in the setting the inner loop actually runs in. k2 is unbiased on-policy, but becomes biased once samples are drawn from pi_theta_old, and needs an importance weight to recover. k1 and k3 are biased in both regimes: k1's gradient vanishes in expectation, and k3's gradient descends the forward KL rather than the reverse one. (see the maths below)

## Fix

Add an opt-in flag `--algo.kl.unbiased_gradient` (default `False`, so existing runs are identical). When enabled, `compute_approx_kl` keeps the chosen estimator's forward value (for logging / reward continuity) but backpropagates
through the correct gradient estimator `w * k1 * grad log pi_theta`, where `w = pi_theta / pi_theta_old` is the importance weight (clamped and detached). See the proof below that this is unbiased.

## Relation to #1101
  This builds on #1101 (which fixes the on-policy gradient bias) by adding the missing off-policy importance weight.
  
## Tests

  `tests/test_kl_estimator_gradient.py` asserts:
  - flag off is identical to the legacy estimator (regression guard);
  - with the flag on, k1/k2/k3 produce the **same** gradient, equal to the analytic
    reverse-KL gradient (on-policy);
  - the gradient scales by the clamped importance weight off-policy;
  - the straight-through preserves each estimator's forward value
  
## Maths

maths

### DIFF
--- a/openrlhf/cli/train_ppo_ray.py
+++ b/openrlhf/cli/train_ppo_ray.py
@@ -427,6 +427,16 @@ if __name__ == "__main__":
             "In GRPO, k3 is utilized as the loss function, while k2, when used as the loss, is nearly equivalent to k1."
         ),
     )
+    parser.add_argument(
+        "--algo.kl.unbiased_gradient",
+        action="store_true",
+        default=False,
+        help=(
+            "When using KL as a loss, keep the chosen estimator's value but backprop the true reverse-KL "
+            "gradient with an importance-sampling correction (w = pi_theta / pi_theta_old) for the off-policy "
+            "inner loop. With this on, k1/k2/k3 share the same unbiased gradient. Default off = legacy behavior."
+        ),
+    )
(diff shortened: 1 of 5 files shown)


anas353:
@LiyuanLucasLiu this builds on #1101 by adding the off-policy importance weight — happy to adjust if you see a cleaner way to combine the two.

LYMDLUT (approved):

PR #1313, opened 2026-08-21
fix(data): filter preference pairs collapsed by truncation

### LINKED ISSUE #1311: [Bug] RewardDataset silently keeps preference pairs that become identical after truncation
## Summary

`RewardDataset` independently truncates the chosen and rejected sequences to `max_length`, then overwrites the final retained token with EOS. If the preference-defining difference lies after the truncation boundary—or exactly at the final retained token—the two model inputs become identical. The dataset still keeps and trains on that pair.

For reward modeling and DPO, identical chosen/rejected inputs carry no preference signal: the pair loss is `ln(2)` and the main preference gradient is exactly zero. In mixed batches, these constant-loss samples also dilute the gradient and distort accuracy/step accounting.

**Suggested severity: medium-high.** The bug is reachable through both public RM and DPO CLIs, whose default `--data.max_len` is 512. Its likelihood depends on the response-length distribution; the shipped shell recipes explicitly use 8192, which substantially reduces it.

Tested on `main` at `3c3be6234e0cb353e76bb8019947db9dfe99fca7` (`v0.11.0`).

## Root cause

The dataset's preprocessing filter only rejects `prompt is None`, plus an independent prompt-length check for DPO:

- [`RewardDataset.__init__`](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/datasets/reward_dataset.py#L87-L99)
- [`RewardDataset.process_data`](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/datasets/reward_dataset.py#L101-L132)

At retrieval time, chosen and rejected text are independently right-truncated and their final retained tokens are forcibly replaced with EOS:

- [`RewardDataset.__getitem__`](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/datasets/reward_dataset.py#L138-L177)

There is no validation after these transformations. The DPO prompt-length filter does not help when the prompt fits but the answers share a long prefix. RM has no equivalent prompt-length filter at all.

The affected losses depend on chosen-minus-rejected values:

- [`PairWiseLoss`](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/models/loss.py#L272-L284)
- [`DPOLoss`](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/models/loss.py#L300-L333)

Identical inputs therefore make the preference difference zero as a function of the model parameters, producing `loss = ln(2)` and zero main preference gradient.

## Stable reproduction with Qwen3.5

Run from the repository root. `MODEL_PATH` may be a local Qwen3.5-0.8B directory or the corresponding Hub ID.

```python
import torch
import torch.nn.functional as F
from datasets import Dataset
from types import SimpleNamespace as NS
from transformers import AutoTokenizer

from openrlhf.datasets import RewardDataset

MODEL_PATH = "/path/to/Qwen3.5-0.8B"
tokenizer = AutoTokenizer.from_pretrained(
    MODEL_PATH,
    trust_remote_code=True,
    local_files_only=True,
)

common = "shared token sequence " * 80
raw = Dataset.from_list(
    [
        {
            "prompt": "Question: ",
            "chosen": common + "CORRECT",
            "rejected": common + "WRONG",
        }
    ]
)
strategy = NS(
    args=NS(
        data=NS(
            prompt_key="prompt",
            chosen_key="chosen",
            rejected_key="rejected",
            apply_chat_template=False,
        )
    )
)

for is_dpo in (False, True):
    dataset = RewardDataset(
        raw,
        tokenizer,
        max_length=32,
        strategy=strategy,
        is_dpo=is_dpo,
        num_processors=None,
    )
    chosen_ids, _, rejected_ids, _, _ = dataset[0]
    print("DPO" if is_dpo else "RM", "identical:", torch.equal(chosen_ids, rejected_ids))

x = torch.tensor(0.0, requires_grad=True)
rm_loss = -F.logsigmoid(x - x)
rm_loss.backward()
print("RM loss/grad:", rm_loss.item(), x.grad.item())

x = torch.tensor(0.0, requires_grad=True)
dpo_loss = -F.logsigmoid(0.1 * ((x - x) - (0.0 - 0.0)))
dpo_loss.backward()
print("DPO loss/grad:", dpo_loss.item(), x.grad.item())
```

Observed on unpatched `main`:

```text
RM identical: True
DPO identical: True
RM loss/grad: 0.6931471824645996 0.0
DPO loss/grad: 0.6931471824645996 0.0
```

The untruncated chosen and rejected sequences are different; the only distinction is beyond the retained token budget.

There is a second boundary case: if the differing token is the last retained token, the explicit EOS overwrite also removes that distinction even though truncation initially retained it.

## CLI reachability and practical relevance

Both public CLIs construct `RewardDataset` directly, and both default to `--data.max_len 512`:

- [RM CLI](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/cli/train_rm.py#L285-L298)
- [DPO CLI](https://github.com/OpenRLHF/OpenRLHF/blob/3c3be6234e0cb353e76bb8019947db9dfe99fca7/openrlhf/cli/train_dpo.py#L305-L318)

As a bounded relevance check—not a full prevalence estimate—I ran the real preprocessing and Qwen3.5 tokenizer over 10 deterministic pages of 100 rows from the public `OpenRLHF/preference_dataset_mixture2_and_safe_pku` dataset. At `max_length=512`, 97 of 1,000 sampled rows produced identical final RM and DPO pairs (only two were already identical before truncation). At 8192, none of those sampled rows collapsed. This is why the severity is medium-high rather than high and why the shipped 8192-token example commands should not be described as always affected.

## Expected behavior

Preference pairs whose final `input_ids` and `attention_mask` are identical should be filtered before dataloader construction/sharding. Pairs whose distinction remains inside the retained token budget must be kept.

Changing global truncation direction is not necessary and could alter training semantics for valid data. Failing later in the collator or loss would also leave dataset length and distributed sharding inconsistent.

## Suggested fix

Reuse the exact tokenization/EOS transformation for preprocessing and `__getitem__`, then set `prompt=None` when the final chosen/rejected `input_ids` and masks are equal. The existing dataset filter already removes `prompt=None` rows.

This adds one-time pair validation during dataset construction while leaving the model input format unchanged. It also avoids duplicating the EOS/truncation logic between validation and retrieval.

## Duplicate check

I searched all open and closed issues and all PRs for combinations of `RewardDataset truncation`, `chosen rejected identical`, `same input preference`, `zero gradient DPO truncation`, and related terms. I found no report or fix for post-truncation pair collapse.

The closest items are not duplicates:

- [#409](https://github.com/OpenRLHF/OpenRLHF/issues/409) is an old preprocessing-speed request that led to parallel `dataset.map`; it does not validate final pair equality.
- [#1025](https://github.com/OpenRLHF/OpenRLHF/issues/1025) concerns packed-sequence truncation in `unpad_and_slice_tensor`, not RM/DPO preference pairs.
- [#909](https://github.com/OpenRLHF/OpenRLHF/issues/909) concerns chat-template/prompt formatting.
- Open [#1247](https://github.com/OpenRLHF/OpenRLHF/pull/1247) adds an optional REBEL-style loss and margin plumbing. It still independently truncates chosen/rejected inputs and contains no equality filter. It may require a straightforward rebase because it touches the same file, but it is not a semantic duplicate.

ai-yang:
## Summary

- filter RM/DPO pairs whose chosen and rejected model inputs become identical after truncation and EOS replacement
- reuse one tokenization path for preprocessing validation and `__getitem__`
- add RM/DPO regression coverage for differences beyond truncation, differences overwritten at the EOS boundary, and valid retained differences

Closes #1311

## Root cause and fix

`RewardDataset` previously filtered long DPO prompts but never checked the final pair. Chosen and rejected sequences were independently truncated, then their last retained tokens were overwritten with EOS. A long shared answer prefix could therefore erase the only preference-defining difference while the sample remained in the dataset.

The patch extracts the existing text cleanup, tokenization, truncation, and EOS overwrite into `_tokenize()`. `process_data()` uses that exact path to compare the final chosen/rejected `input_ids` and `attention_mask`; identical pairs reuse the existing `prompt=None` filter. `__getitem__()` uses the same helper, so validation cannot drift from the tensors consumed by RM/DPO training.

Dataset construction now performs one-time tokenization of both sides for validation. This is intentional: filtering in `__getitem__`, the collator, or the loss would leave dataset length and distributed sharding inconsistent. The runtime model-input format is unchanged, and the helper reduces duplicated production logic (28 added / 29 removed lines in `reward_dataset.py`).

## Tests

- fail-before: `tests/test_reward_dataset.py` produced `4 failed, 2 passed` on unpatched `main`
- pass-after: targeted tests — `6 passed`
- full test suite — `28 passed`
- real Qwen3.5-0.8B tokenizer check:
  - unpatched RM/DPO: dataset length `1`, chosen/rejected IDs identical
  - patched RM/DPO: dataset length `0`, collapsed pair filtered
- `python -m compileall -q openrlhf examples tests`
- project pre-commit hooks on the changed files
- `bash -n` on all runnable example scripts

The tests are parameterized over RM and DPO and cover:

1. the only difference lies beyond `max_length`;
2. the difference is the final retained token and is overwritten by EOS;
3. the difference remains inside the retained range and the pair is kept.

`examples/scripts/train_ppo_ray_slurm.sh` was excluded from the shell check because its literal `<OPENRLHF_ROOT_PATH>` template placeholder already fails `bash -n` unchanged on `main`.

### DIFF
--- a/openrlhf/datasets/reward_dataset.py
+++ b/openrlhf/datasets/reward_dataset.py
@@ -98,6 +98,24 @@ class RewardDataset(Dataset):
         self.rejects = processed_dataset["reject"]
         self.extras = processed_dataset["extra"]
 
+    def _tokenize(self, text):
+        text = text.rstrip("\n")
+        if not text.endswith(self.tokenizer.eos_token):
+            text += " " + self.tokenizer.eos_token
+        token = self.tokenizer(
+            text,
+            max_length=self.max_length,
+            padding=False,
+            truncation=True,
+            return_tensors="pt",
+            add_special_tokens=False,
+        )
+
+        # to avoid EOS_token truncation
+        token["input_ids"][0][-1] = self.tokenizer.eos_token_id
+        token["attention_mask"][0][-1] = True
+        return token
+
     def process_data(self, data):
         prompt, chosen, reject, margin = preprocess_data(
             data,
@@ -124,6 +142,14 @@ class RewardDataset(Dataset):
             if prompt_ids_len >= self.max_length - 2:
                 prompt = None
 
+        if prompt is not None:
+            chosen_token = self._tokenize(prompt + chosen)
+            reject_token = self._tokenize(prompt + reject)
+            same_input_ids = chosen_token["input_ids"].equal(reject_token["input_ids"])
+            same_attention_mask = chosen_token["attention_mask"].equal(reject_token["attention_mask"])
+            if same_input_ids and same_attention_mask:
+                prompt = None
+
         return {
             "prompt": prompt,
             "chosen": chosen,
@@ -138,35 +164,8 @@ class RewardDataset(Dataset):
     def __getitem__(self, idx):
         prompt, chosen, reject, extra = self.prompts[idx], self.chosens[idx], self.rejects[idx], self.extras[idx]
 
-        chosen = (prompt + chosen).rstrip("\n")
-        if not chosen.endswith(self.tokenizer.eos_token):
-            chosen += " " + self.tokenizer.eos_token
-        chosen_token = self.tokenizer(
-            chosen,
-            max_length=self.max_length,
-            padding=False,
-            truncation=True,
-            return_tensors="pt",
-            add_special_tokens=False,
-        )
-
-        reject = (prompt + reject).rstrip("\n")
-        if not reject.endswith(self.tokenizer.eos_token):
-            reject += " " + self.tokenizer.eos_token
-        reject_token = self.tokenizer(
-            reject,
-            max_length=self.max_length,
-            padding=False,
-            truncation=True,
-            return_tensors="pt",
-            add_special_tokens=False,
-        )
-
-        # to avoid EOS_token truncation
-        chosen_token["input_ids"][0][-1] = self.tokenizer.eos_token_id
-        reject_token["input_ids"][0][-1] = self.tokenizer.eos_token_id
-        chosen_token["attention_mask"][0][-1] = True
-        reject_token["attention_mask"][0][-1] = True
+        chosen_token = self._tokenize(prompt + chosen)
+        reject_token = self._tokenize(prompt + reject)
 
         return (
             chosen_token["input_ids"],
--- a/openrlhf/datasets/sft_dataset.py
+++ b/openrlhf/datasets/sft_dataset.py
@@ -6,9 +6,7 @@ from torch.utils.data import Dataset
 from openrlhf.utils.utils import zero_pad_sequences
 
 
-def preprocess_data(
-    data, input_template=None, input_key="input", output_key=None, apply_chat_template=None
-):
+def preprocess_data(data, input_template=None, input_key="input", output_key=None, apply_chat_template=None):
     if apply_chat_template:
         if output_key:
             prompt_message = data[input_key]
--- a/tests/test_reward_dataset.py
+++ b/tests/test_reward_dataset.py
@@ -1,0 +1,62 @@
+from types import SimpleNamespace
+
+import pytest
+import torch
+from datasets import Dataset
+
+from openrlhf.datasets import RewardDataset
+
+
+class _Tokenizer:
+    eos_token = "<eos>"
+    eos_token_id = 99
+    pad_token_id = 0
+
+    def __call__(self, text, max_length, **_kwargs):
+        token_ids = {
+            "prompt": 1,
+            "common": 2,
+            "chosen": 3,
+            "rejected": 4,
+            self.eos_token: self.eos_token_id,
+        }
+        input_ids = torch.tensor([[token_ids[token] for token in text.split()[:max_length]]])
+        return {"input_ids": input_ids, "attention_mask": torch.ones_like(input_ids)}
+
+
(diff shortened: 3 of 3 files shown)


Excelius-Wang (commented):
Comparing the final tokenized inputs is the right boundary here: replacing the last retained token with EOS can erase a preference even when truncation alone did not.

I reproduced both collapse cases in RM and DPO modes and also checked them with a locally constructed Hugging Face tokenizer rather than the test stub. Collapsed pairs are removed during dataset construction, while differences inside the retained tokens are kept. The CPU suite and additional tokenizer checks pass. I didn't run model training or benchmark the extra preprocessing cost.

PR #1323, opened 2026-09-04
fix(async): persist oversampling buffer state across checkpoints

### LINKED ISSUE #1322: [Bug] Async oversampling checkpoints drop buffered rollouts on resume
# [Bug] Async oversampling checkpoints drop buffered rollouts on resume

## Severity

Medium-high. A normal checkpoint reports success and restores model/optimizer/dataloader state, but silently skips generated prompts that had not yet been trained.

## Affected revision

`main` at `3c3be6234e0cb353e76bb8019947db9dfe99fca7` (v0.11.0).

## Preconditions

- `--train.async_enable`
- `--rollout.vllm_generate_batch_size > --rollout.batch_size`

This is the documented oversampling configuration in the README.

## Summary

`SamplesGenerator.generate_samples()` advances the stateful prompt dataloader by the full vLLM generation batch and keeps extra `Experience` objects in `_sample_buffer`. Async checkpoint client state saves only `prompts_dataloader.state_dict()`. On restart the dataloader resumes after every generated prompt, while `_sample_buffer` is initialized empty, so all generated-but-untrained rollouts are permanently skipped.

The checkpoint also lacks an explicit dataloader-exhausted bit. Recreating an iterator from a finished `StatefulDataLoader` can start the same episode again, especially when the dataset ends exactly on a generation-batch boundary and no final `StopIteration` probe was needed.

## Minimal reproduction

Use a stateful loader containing IDs `0..7`, `rollout.batch_size=2`, `n_samples_per_prompt=1`, and `vllm_generate_batch_size=4`:

1. The first generation advances the loader through ID 3.
2. The first training chunk is `[0, 1]`; `_sample_buffer` retains `[2, 3]`.
3. Save the current loader state (`position=4`) as async training does.
4. Construct a new generator, load that state, and generate the next chunk.

Current result:

```text
first_trained_chunk=[0, 1]
checkpointed_loader_position=4
live_buffer_at_checkpoint=[2, 3]
checkpointed_buffer=[]
first_chunk_after_resume=[4, 5]
```

Expected `first_chunk_after_resume` is `[2, 3]`. [PR #1323's committed regression suite](https://github.com/ai-yang/OpenRLHF/blob/015d391fd234a6f255675f2d47d673c6b6e4d8e7/tests/test_ppo_trainer.py) covers buffer capture, checkpoint materialization, restore, exact boundaries, and exhausted-buffer variants.

## Root cause

The generator sends this checkpoint state with each produced rollout:

```python
client_states = {
    "episode": episode,
    "total_consumed_prompts": total_consumed_prompts,
    "data_loader_state_dict": self.prompts_dataloader.state_dict(),
}
```

However, the state required to continue from that exact training boundary is the dataloader state, the remaining `_sample_buffer`, and whether the current episode's dataloader is already exhausted. `GenerateSamplesActor.load_state_dict()` restores only the loader cursor.

Because async generation can run ahead of training, checkpoint state must stay associated with each queued rollout. Reading the generator's current buffer only when the trainer saves would capture a later boundary.

## Impact

- Resume silently changes the training data stream instead of failing.
- Up to almost `vllm_generate_batch_size * n_samples_per_prompt` rollouts can be lost at one checkpoint boundary.
- If the oversampled generation drained the dataset, a resumed episode can finish immediately even though most generated samples were never optimized.
- At an exact generation boundary, recreating the finished loader can instead repeat already-trained prompts from the same episode.
- Prompt accounting still includes skipped prompts, so logs do not expose the loss.

## Duplicate search

I searched all issue and PR states for `oversampling`, `vllm_generate_batch_size`, `sample_buffer`, `async checkpoint`, `resume`, `dataloader`, and `skipped samples`.

- Merged #1152 introduced the current async sampler/checkpoint restore path, but its client state contains only the loader cursor and prompt counters.
- #1103 and merged #1124 concern progress-bar display after a correctly restored dataloader; the reporter confirmed that the loader state itself did load.
- Open #1179 proposes an earlier oversampling design but does not persist the current `_sample_buffer` state.
- #1297 / merged #1300 fixes a non-empty terminal batch being discarded during uninterrupted synchronous training, not state lost across async checkpoint restore.

No existing issue or PR covers loss of the current `_sample_buffer` across async checkpoint restore.

ai-yang:
# fix(async): persist oversampling buffer state across checkpoints

Fixes #1322

The terminal-tail prerequisite #1321 has merged. This PR contains the checkpoint-specific follow-up.

## Current status and latest validation (2026-09-21)

The terminal-tail prerequisite #1321 has merged. This branch now incorporates main `dc2a7ad`; conflicts in the two test files were resolved by retaining the upstream regressions alongside the checkpoint regressions. The diff against main contains only the checkpoint change in `ppo_trainer.py`, `ppo_trainer_async.py`, and `test_ppo_trainer.py`.

- Targeted trainer/sample-generator tests: **19 passed**.
- Complete CPU suite: **211 passed, 2 skipped, 14 warnings**, using real PyTorch 2.13.0 and the existing import-only DeepSpeed shim. The optional dependency skips are not evidence of distributed training validation.
- Black, isort, `compileall` and `git diff --check`: passed.
- `bash -n`: 17 example scripts pass. The unchanged upstream SLURM template has the literal placeholder `OPENRLHF_PATH=<OPENRLHF_ROOT_PATH>` and needs configuration before parsing.

The earlier pinned Ray/DeepSpeed boundary validation below remains historical evidence; it was not rerun as part of this test-only conflict resolution. OpenAI Codex assisted with this update.

## What changed

- Associate each queued rollout with the generated-but-untrained buffer suffix at that exact training boundary.
- Put each newly generated tail in Ray's object store once; later chunks share that reference and record only the remaining suffix length.
- Materialize the suffix only when writing a regular or best checkpoint, avoiding overlapping copies in every queue payload.
- Persist whether the current episode's prompt loader is exhausted, including exact generation-batch boundaries.
- Restore the buffer, loader cursor, and exhaustion state before async generation resumes.
- Keep checkpoints created before this change loadable when the new fields are absent.

## Why

With `vllm_generate_batch_size > rollout.batch_size`, generation advances the prompt dataloader by the larger generation batch but trains only one rollout-sized chunk. The remaining generated experiences live in `_sample_buffer`.

The checkpoint previously stored the already-advanced dataloader cursor but not that buffer. A minimal generation-batch 4 / rollout-batch 2 run trained `[0, 1]`, saved loader position 4, and resumed at `[4, 5]`; buffered rollouts `[2, 3]` were silently lost.

An explicit exhaustion bit is also needed. `StatefulDataLoader` does not report exhaustion until another read is attempted, so a dataset ending exactly at a generation boundary otherwise checkpoints the final trained state as non-exhausted and can restart the same episode after resume.

## Design and compatibility

- Async generation may be ahead of training, so each queue item must retain its own checkpoint boundary.
- A nested Ray `ObjectRef` keeps the immutable generated tail alive without serializing the same `Experience` objects into every overlapping queue state.
- Buffer-only calls remove a prefix, so the remaining length identifies the exact suffix. Any call that consumes new prompts creates a fresh object-store snapshot.
- Only persistent checkpoints contain the materialized `sample_buffer`; transient ObjectRefs are never written to disk.
- Old checkpoints default to an empty buffer and non-exhausted loader. They remain loadable, although data already omitted by an old checkpoint cannot be reconstructed.

## Initial validation (2026-09-04)

```text
python -m pytest -q tests/test_ppo_trainer.py
10 passed

PYTHONPATH=../OpenRLHF-audit-report-20260904/test_stubs python -m pytest -q
31 passed

python -m compileall -q openrlhf examples tests
passed
```

The full-suite command used this import-only stub because DeepSpeed is not installed in the local environment:

```python
class _Utils:
    @staticmethod
    def set_z3_leaf_modules(*args, **kwargs):
        pass


utils = _Utils()
```

New regression coverage verifies:

- old checkpoint defaults and non-exhausted restore;
- one shared snapshot across multiple queue states, with exact suffix recovery;
- exhausted restore with empty, partial, exact, and multiple full chunks;
- exact generation-boundary exhaustion across every queued state;
- lazy materialization for regular checkpoints and best checkpoints;
- combined behavior with the terminal-tail fix.

`git diff --check`, isort, and Black's formatting API passed for the PR diff. No GPU is required because this is deterministic controller and checkpoint-state logic.

The earlier bot-authored formatting change is already present upstream and is no longer part of this PR diff.

## Pinned-dependency runtime validation

The committed unit tests isolate controller behavior with lightweight mocks. I additionally validated the external boundaries using the repository's exact Ray 2.55.0 dependency and the pinned DeepSpeed 0.19.5 checkpoint implementation:

- A generator actor created the buffer `ObjectRef`, placed it as a nested value in `ray.util.queue.Queue`, dropped its local Python reference, and a different actor resolved the same object ID and suffix (`same_id=True`, `remaining=[4, 5]`).
- A real `Experience` containing CPU tensors survived an on-disk `torch.save` / `torch.load(weights_only=False)` client-state round trip. DeepSpeed 0.19.5's `TorchCheckpointEngine.load()` explicitly uses that load mode.

The complete smoke script, output, and pinned DeepSpeed source link are in [this validation comment](https://github.com/OpenRLHF/OpenRLHF/pull/1323#issuecomment-5542686112).

## Checklist

- [x] The committed regression suite fails on the affected base revision and passes on this branch.
- [x] One generated tail is stored once rather than copied into every queue payload.
- [x] Persistent checkpoints contain no Ray ObjectRefs.
- [x] Exact-boundary and exhausted-buffer restore cases are covered.
- [x] Full unit suite and syntax compilation pass.

### DIFF
--- a/openrlhf/datasets/sft_dataset.py
+++ b/openrlhf/datasets/sft_dataset.py
@@ -6,9 +6,7 @@ from torch.utils.data import Dataset
 from openrlhf.utils.utils import zero_pad_sequences
 
 
-def preprocess_data(
-    data, input_template=None, input_key="input", output_key=None, apply_chat_template=None
-):
+def preprocess_data(data, input_template=None, input_key="input", output_key=None, apply_chat_template=None):
     if apply_chat_template:
         if output_key:
             prompt_message = data[input_key]
--- a/openrlhf/trainer/ppo_trainer.py
+++ b/openrlhf/trainer/ppo_trainer.py
@@ -362,6 +362,10 @@ class BasePPOTrainer(ABC):
             client_states["best_eval_metric_key"] = metric_key
             client_states["best_eval_metric_value"] = current_value
             client_states["checkpoint_metric_key"] = metric_key
+            sample_buffer_state = client_states.pop("sample_buffer_state", None)
+            if sample_buffer_state:
+                sample_buffer_ref, size = sample_buffer_state
+                client_states["sample_buffer"] = ray.get(sample_buffer_ref)[-size:]
 
             tag = f"best_global_step{global_step}"
             refs = self.actor_model_group.async_run_method(
@@ -432,7 +436,9 @@ class BasePPOTrainer(ABC):
             checkpoint_states = ray.get(self.actor_model_group.async_run_method(method_name="get_checkpoint_states"))[
                 0
             ]
-            logger.info(f"checkpoint_states: {checkpoint_states}")
+            checkpoint_log = dict(checkpoint_states)
+            checkpoint_log["sample_buffer"] = len(checkpoint_log.get("sample_buffer", []))
+            logger.info(f"checkpoint_states: {checkpoint_log}")
             return checkpoint_states
         return {
             "episode": 0,
--- a/openrlhf/trainer/ppo_trainer_async.py
+++ b/openrlhf/trainer/ppo_trainer_async.py
@@ -62,6 +62,8 @@ class GenerateSamplesActor:
             tokenizer=tokenizer,
             vllm_engines=vllm_engines,
         )
+        self._sample_buffer_ref = None
+        self._dataloader_exhausted = False
 
         self.vllm_lock = vllm_lock
         self._partial_rollout = getattr(strategy.args.train, "partial_rollout_enable", False)
@@ -73,8 +75,12 @@ class GenerateSamplesActor:
     def get_max_steps(self):
         return self.max_steps
 
-    def load_state_dict(self, state_dict):
+    def load_state_dict(self, state_dict, sample_buffer=None, dataloader_exhausted=False):
         self.prompts_dataloader.load_state_dict(state_dict)
+        self.samples_generator._sample_buffer = list(sample_buffer or [])
+        self.samples_generator._dataloader_iter = iter(()) if dataloader_exhausted else iter(self.prompts_dataloader)
+        self._dataloader_exhausted = dataloader_exhausted
+        self._sample_buffer_ref = ray.put(self.samples_generator._sample_buffer) if sample_buffer else None
 
     def _should_eval(self, global_step):
         return (
@@ -139,12 +145,33 @@ class GenerateSamplesActor:
                     if not self._partial_rollout:
                         ray.get(self.vllm_lock.release.remote())
 
+                if (
+                    self.samples_generator._dataloader_iter is None
+                    or total_consumed_prompts >= (episode + 1) * dataset_length
+                ):
+                    self._dataloader_exhausted = True
+                elif prompts_consumed:
+                    self._dataloader_exhausted = False
+
+                # Queue entries share one generated tail and identify its remaining suffix.
+                # This avoids serializing overlapping Experience lists for every training chunk.
+                if prompts_consumed and self.samples_generator._sample_buffer:
+                    self._sample_buffer_ref = ray.put(self.samples_generator._sample_buffer)
+                elif not self.samples_generator._sample_buffer:
+                    self._sample_buffer_ref = None
+
                 produced = bool(rollout_samples)
                 if produced:
                     client_states = {
                         "episode": episode,
                         "total_consumed_prompts": total_consumed_prompts,
                         "data_loader_state_dict": self.prompts_dataloader.state_dict(),
+                        "dataloader_exhausted": self._dataloader_exhausted,
+                        "sample_buffer_state": (
+                            (self._sample_buffer_ref, len(self.samples_generator._sample_buffer))
+                            if self._sample_buffer_ref is not None
+                            else None
+                        ),
                     }
                     self.rollout_queue.put(
                         (rollout_samples, client_states, filter_pass_rate, generation_time), block=True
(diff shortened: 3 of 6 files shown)


ai-yang:
Additional runtime validation for the two external boundaries that the committed unit tests mock: Ray object ownership and on-disk checkpoint serialization.

### Ray 2.55.0 actor-to-actor Queue smoke

The repository pins `ray[default]==2.55.0`. I ran this with that exact version:

```python
import gc

import ray
from ray.util.queue import Queue


@ray.remote
class Producer:
    def enqueue(self, queue):
        ref = ray.put([2, 3, 4, 5])
        ref_id = ref.hex()
        queue.put({"sample_buffer_state": (ref, 2)})
        del ref
        gc.collect()
        return ref_id


@ray.remote
class Consumer:
    def restore(self, queue):
        payload = queue.get()
        ref, size = payload["sample_buffer_state"]
        return ref.hex(), ray.get(ref)[-size:]


ray.init(num_cpus=2, include_dashboard=False, log_to_driver=False)
queue = Queue(maxsize=1)
producer = Producer.remote()
consumer = Consumer.remote()
expected_id = ray.get(producer.enqueue.remote(queue))
actual_id, remaining = ray.get(consumer.restore.remote(queue))
assert actual_id == expected_id
assert remaining == [4, 5]
queue.shutdown()
ray.shutdown()
```

Result:

```text
ray_version=2.55.0
same_id=True
remaining=[4, 5]
```

This exercises the production ownership topology: the generator actor creates the `ObjectRef`, puts it as a nested value in `ray.util.queue.Queue`, drops its local Python reference, and a different actor resolves the same object ID.

### DeepSpeed client-state round trip

I also saved a real `Experience` containing CPU tensors inside `{"sample_buffer": [...]}` with `torch.save`, then loaded it with `weights_only=False`; the restored object retained its type, tensors, and prompt metadata.

This matches the pinned DeepSpeed 0.19.5 checkpoint path: [`TorchCheckpointEngine.load()` explicitly calls `torch.load(..., weights_only=False)`](https://github.com/deepspeedai/DeepSpeed/blob/cf44300453eb0af79ed84ed8f1cb49d57478bd76/deepspeed/runtime/checkpoint_engine/torch_checkpoint_engine.py#L35-L39). Before persistence, the PR removes `sample_buffer_state`, resolves its nested Ray reference, and writes only the materialized `sample_buffer`, so no `ObjectRef` reaches disk.

These are CPU-only state/serialization checks; model execution and GPU kernels are not involved in either failure.

PR #1324, opened 2026-09-06
fix(data): retain trainable turns before a truncated prompt

feng1201:
With `data.multiturn=true`, an SFT conversation can contain trainable assistant answers before a long final prompt. The dataset currently applies the final prompt's length cutoff to the whole row, dropping those earlier answers as well. Check whether any assistant span survives the existing cutoff in multi-turn mode; retain the existing single-turn filtering rule.

The regression uses the actual SFTDataset constructor, item access and collator with a real Hugging Face dataset and locally constructed tokenizer. It covers long/short final prompts, with/without an earlier answer, and single/multi-turn modes, and checks which target tokens receive loss.

Validation (CPU, Python 3.11.15, Torch 2.13.0, Transformers 5.15.0):

- Original regression: **1 failed, 7 passed**. A row with a usable earlier answer was incorrectly removed only in the long-final-prompt multi-turn case.
- Fixed regression: **8 passed**.
- Full repository pytest suite: **30 passed**, 14 warnings, no skips.
- `python -m compileall -q openrlhf examples tests`: passed.
- All applicable configured pre-commit hooks on the changed files: passed.

```bash
python -m pytest -q
python -m compileall -q openrlhf examples tests
pre-commit run --files openrlhf/datasets/sft_dataset.py tests/test_sft_truncation.py
```

No GPU training was run. DeepSpeed 0.19.5 was installed with native-op compilation disabled to run the existing CPU unit tests. The new dataset regression uses a Namespace only to hold the CLI data configuration; it does not replace dataset/tokenizer logic.

Duplicate checks: reviewed all open PRs and searches for multiturn/filter and response_ranges. #1227 changes Qwen3 thinking-template mask construction and output-field normalization; its patch retains the final-prompt filter and does not fix this row-selection condition.

AI assistance: OpenAI Codex assisted with investigation, implementation and testing.

### DIFF
--- a/openrlhf/datasets/sft_dataset.py
+++ b/openrlhf/datasets/sft_dataset.py
@@ -6,9 +6,7 @@ from torch.utils.data import Dataset
 from openrlhf.utils.utils import zero_pad_sequences
 
 
-def preprocess_data(
-    data, input_template=None, input_key="input", output_key=None, apply_chat_template=None
-):
+def preprocess_data(data, input_template=None, input_key="input", output_key=None, apply_chat_template=None):
     if apply_chat_template:
         if output_key:
             prompt_message = data[input_key]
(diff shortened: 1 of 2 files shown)


Excelius-Wang (commented):
This makes sense: a truncated final prompt shouldn't discard earlier assistant turns that still fit. The single-turn filtering rule stays unchanged.

I reproduced the dropped-row case with the old code. With this change, I also checked the shifted loss mask near the cutoff: it selects the earlier answer and its EOS, without training on the following user prompt. The CPU suite and additional cutoff checks pass; I haven't run training.

